PR #12, opened 2014-02-16
Set headers property on WebSocketServerProtocol

maximesteisel:
Having access to headers from ws_handler is handy to do Origin filtering. 

It seems logical to put it in the protocol since there is only one handshake per connection, and the handshake is guaranteed to be finished before calling ws_handler()


aaugustin:
Related: #494.

Issue #38, opened 2014-11-23
Allow Sub-Protocol negotiation during the Websocket handshake.


eisensheng:
The [RFC6455 Section 1.9](http://tools.ietf.org/html/rfc6455#section-1.9) leaves us with a header field to negotiate a sub-protocol during the Websocket handshake. It would be nice if you could allow your library users to interfere at this time so we can see which protocols are advertised by the client, choose one of them or enforce a completely different value and send it back to the client. Of course the ``ws_handler`` needs to know which protocol has been chosen.

aaugustin:
Can you have a look at this pull request and tell me if it works for you? Thanks!

eisensheng:
Haven't tested it but the PR looks amazing! I especially like the way one can override `select_subprotocol` to interfere at this point while the default implementation already looks perfect. Thank you very much for this one :+1:

PR #43, opened 2015-01-31
Support non-default event loop.

### LINKED ISSUE #42: Allow to pass event loop into websockets.serve()
```websockets``` does not currently allows us to explicitly pass an event loop instance into ```websockets.serve()```. Subsequent call to ```create_server()``` relies on existence of the default event loop, which is, however, not guaranteed, even in the main thread, since the calling code may deliberately suppress invoking of default event loop. According to PEP 3156:
> It is okay to set the current event loop to None , in which case subsequent calls to get_event_loop() will raise an exception. This is useful for testing code that should not depend on the existence of a default event loop.

Also see  https://www.python.org/dev/peps/pep-3156/#passing-an-event-loop-around-explicitly  

Most of ```asyncio``` calls allow us to explicitly pass an event loop instance via the keyword-only ```loop``` argument. Please consider implementing such argument on ```websockets.serve()``` to benefit the code that not relies on the existence of the default event loop.

aaugustin:
Fix #42.

### DIFF
--- a/docs/index.rst
+++ b/docs/index.rst
@@ -117,7 +117,7 @@ Server
 
 .. automodule:: websockets.server
 
-   .. autofunction:: serve(ws_handler, host=None, port=None, *, klass=WebSocketServerProtocol, origins=None, **kwds)
+   .. autofunction:: serve(ws_handler, host=None, port=None, *, loop=None, klass=WebSocketServerProtocol, origins=None, **kwds)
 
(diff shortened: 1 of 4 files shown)


rielsi:
Works perfectly! Thank you very much.

aaugustin:
Thanks. I will push a release to PyPI after working on the other currently open issue (#38).

PR #50, opened 2015-04-03
Timeout on websocket.recv()

Natim:
They are no timeout argument on ``websocket.recv()`` also when using it with:

    yield from asyncio.wait_for(websocket.recv(), 5)

The task is canceled but the next message on the websocket is not received by the next ``websocket.recv()`` call.


Natim:
With some help from @haypo I managed to add a test and fix it.

aaugustin:
Looks great, thanks!

PR #55, opened 2015-05-23
Fix #53 - warnings in example code

### LINKED ISSUE #53: Warning when running the example code on Python 3.4
What's your name? mYk
> mYk
< Hello mYk!
Task was destroyed but it is pending!
task: <Task pending coro=<run() running at /Users/myk/Documents/dev/websockets/websockets/protocol.py:253> wait_for=<Future pending cb=[Task._wakeup()]>>

nolar:
It is quite unclear why client cannot use

```
yield from websocket.close()
```

as it is said in the documentation (“Clients shouldn’t close the WebSocket connection. Instead, they should wait until the server performs the closing handshake by yielding from the protocol’s worker attribute.”)

As such, it is replaced with

```
yield from websocket.worker
```

It works and no warnings are shown, but in this form the code is poorly readable and unclear what does this code do.

### DIFF
--- a/example/client.py
+++ b/example/client.py
@@ -11,6 +11,6 @@ def hello():
     print("> {}".format(name))
     greeting = yield from websocket.recv()
     print("< {}".format(greeting))
-    yield from websocket.close()
+    yield from websocket.worker
 
 asyncio.get_event_loop().run_until_complete(hello())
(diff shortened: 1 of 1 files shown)


aaugustin:
Do you think I should add a `wait_close()` method?

nolar:
```
yield from websocket.close()
```

works fine, no warnings are shown (see the first commit in this request).

I'm not sure it should be explicit, though, since I'm just learning asyncio and websocket, and I'm trying to figure out how to use all these libraries.

aaugustin:
Fixed in 9469965a6552c038232ae987e473199451295730.

Issue #47, opened 2015-03-26
Add support for custom HTTP headers in handshake request


aaugustin:
See #46 for a use case.

knutae:
Created a pull request which allows custom HTTP headers in the client.connect() call.

My first fork and pull request ever on github, let me know if I did something wrong. :)

ye:
@aaugustin thanks for creating this library! @knutae thanks for the PR to add this functionality. Another use case for the custom HTTP header support is the ability to retrieve custom HTTP headers that WebSocket client sending in at the time of handshake. I was looking at the code, but I couldn't find a way to get that at version 2.5. Is it possible without modify the source? Can you shed some lights on my request?

PR #57, opened 2015-06-30
Added a headers keyword to client.connect(), allowing websocket clients to specify custom HTTP headers.

knutae:
Suggested fix for https://github.com/aaugustin/websockets/issues/47


aaugustin:
Thanks for providing this patch, it was a good starting point!

I just committed a change which also adds docs and tests, and provides the ability to customize response headers.

Issue #60, opened 2015-07-15
Pass event loop explicitly to all asyncio objects


SzieberthAdam:
Python: 3.4
OS: Windows 7 (Firewall turned off)
Virtual environment was used (pyvenv).

Server code:

```python
#!/usr/bin/env python

import asyncio
import threading
import websockets

@asyncio.coroutine
def hello(websocket, path):
    name = yield from websocket.recv()
    print("< {}".format(name))
    greeting = "Hello {}!".format(name)
    yield from websocket.send(greeting)
    print("> {}".format(greeting))

class ServerThread(threading.Thread):

  def run(self):
    self.loop = asyncio.new_event_loop()
    coro = websockets.serve(hello, 'localhost', 8765,
                            loop=self.loop)
    self.loop.run_until_complete(coro)
    self.loop.run_forever()

ServerThread().start()
```

Exception traceback in client.py (documentation example code):

```
Traceback (most recent call last):
  File "C:\tajf\lib\site-packages\websockets\client.py", line 61, in handshake
    status_code, headers = yield from read_response(self.reader)
  File "C:\tajf\lib\site-packages\websockets\http.py", line 63, in read_response

    status_line, headers = yield from read_message(stream)
  File "C:\tajf\lib\site-packages\websockets\http.py", line 80, in read_message
    start_line = yield from read_line(stream)
  File "C:\tajf\lib\site-packages\websockets\http.py", line 99, in read_line
    line = yield from stream.readline()
  File "c:\Python34_32bit\lib\asyncio\streams.py", line 425, in readline
    yield from self._wait_for_data('readline')
  File "c:\Python34_32bit\lib\asyncio\streams.py", line 393, in _wait_for_data
    yield from self._waiter
  File "c:\Python34_32bit\lib\asyncio\futures.py", line 386, in __iter__
    yield self  # This tells Task to wait for completion.
  File "c:\Python34_32bit\lib\asyncio\tasks.py", line 287, in _wakeup
    value = future.result()
  File "c:\Python34_32bit\lib\asyncio\futures.py", line 275, in result
    raise self._exception
  File "c:\Python34_32bit\lib\asyncio\selector_events.py", line 662, in _read_ready
    data = self._sock.recv(self.max_size)
ConnectionAbortedError: [WinError 10053] An established connection was aborted by the software in your host machine

The above exception was the direct cause of the following exception:

Traceback (most recent call last):
  File "client.py", line 15, in <module>
[...]

PR #61, opened 2015-07-16
(Pull request) Fix for #60

SzieberthAdam:
I hope I found all objects without explicit event loop arguments.
Identical unittest result.


aaugustin:
Thanks a lot! I'll review this as soon as I get the opportunity, perhaps during the week-end.

PR #70, opened 2015-08-25
wait for all websockets to close in wait_close

### LINKED ISSUE #69: Handler for closed connections not allowed to finish
During shutdown I wait for clients to close connections. When this is done, the handler is not allowed to finish.

To trigger the error, run the following script, interrupt it (Ctrl+C) with a connection open, then close the connection.

    #!/usr/bin/env python
    
    import asyncio
    import logging
    import signal
    
    import websockets
    
    logging.getLogger().setLevel(logging.DEBUG)
    
    connectionClosedEvents = []
    
    @asyncio.coroutine
    def echo(websocket, path):
        print("Connection opened.")
        connectionClosedEvent = asyncio.Event()
        connectionClosedEvents.append(connectionClosedEvent)
        s = yield from websocket.recv()
        if s is not None:
            yield from websocket.send(s)
        print("Connection closed.")
        connectionClosedEvent.set()
    
    @asyncio.coroutine
    def shutdown():
        print("Shutdown initialized.")
        print("Waiting for open connections to close ...")
        for e in connectionClosedEvents:
            yield from e.wait()
        print("All connections closed.")
        echoServer.close()
        print("Waiting for server to close ...")
        yield from echoServer.wait_closed()
        print("Server closed.")
        print("Stopping event loop ...")
        eventLoop.stop()
        print("Event loop stopped.")
        print("Shutdown complete.")
    
    eventLoop = asyncio.get_event_loop()
    eventLoop.add_signal_handler(signal.SIGINT, asyncio.async, shutdown())
    echoServer =\
        eventLoop.run_until_complete(websockets.serve(echo, "0.0.0.0", 8765))
    eventLoop.run_forever()

Typical output:

    Connection opened.
    ^CShutdown initialized.
    Waiting for open connections to close ...
    Connection closed.
    All connections closed.
    Waiting for server to close ...
    Server closed.
    Stopping event loop ...
    Event loop stopped.
    Shutdown complete.
    Task was destroyed but it is pending!
    task: <Task pending coro=<handler() running at /tmp/tmp.OhmZJ4uKUJ/venv/lib/python3.4/site-packages/websockets/server.py:77> wait_for=<Future finished result=None>>

The reason for the error is that `websockets.server.connection_lost` removes the connection from `self.websockets`. Thus in `websockets.server.wait_closed` the connection is not waited for:

                yield from asyncio.wait(
                    [ws.handler_task for ws in self.websockets], loop=self.loop)

I will submit a proposed fix for this shortly.

anderslundstedt:
This fixes #69

### DIFF
--- a/websockets/server.py
+++ b/websockets/server.py
@@ -189,6 +189,7 @@ class WebSocketServer(asyncio.AbstractServer):
         self.loop = loop
 
         self.websockets = set()
+        self.open_websockets = set()
 
     def wrap(self, server):
         """
@@ -207,15 +208,16 @@ class WebSocketServer(asyncio.AbstractServer):
 
     def register(self, protocol):
         self.websockets.add(protocol)
+        self.open_websockets.add(protocol)
 
     def unregister(self, protocol):
-        self.websockets.remove(protocol)
+        self.open_websockets.remove(protocol)
 
     def close(self):
         """
         Stop serving and trigger a closing handshake on open connections.
         """
-        for websocket in self.websockets:
+        for websocket in self.open_websockets:
             asyncio.async(websocket.close(1001), loop=self.loop)
         self.server.close()
 


aaugustin:
This patch introduces a memory leak. Since references to the protocol for each websocket connection aren't removed from `self.websockets` anymore, the protocols cannot be garbage-collected. This will lead to unlimited memory consumption and eventually a server crash.

PR #90, opened 2015-12-15
Adding context manager for client connect().

SzieberthAdam:
A pull request from my own fork which is a cleaner way to do my part. I learned this by following python/asyncio development.


SzieberthAdam:
Maybe `_Connect` should get the docstring of `client.connect`.

aaugustin:
I really like how this is shaping up. Good work!

Don't worry about the documentation for now. We have stabilized the target API and that's what matters.

It would be interesting to add tests. If we're using `try:  <Python 3.5 only syntax> except SyntaxError: ...` in `client.py`, we should do the same in `test_client_server.py` to define tests for the new syntax.

`tox.ini` should run the coverage tests under python3.5, otherwise the new syntax won't be tested and tox will fail (it requires 100% coverage). Let's minimize code that's only for python3.4 and 3.3 and skip coverage on that code.

SzieberthAdam:
Haha, now I see that making a PR from an own fork signals and implies that the case is my own responsibility to get solved and covered properly. :)

No problem. This is a good opportunity for me to get more familiar with the tests as well.

aaugustin:
It's up to you! If you want to leave the space for someone else, just say so :-)

Let's just communicate about who does what in order to avoid duplicating work.

SzieberthAdam:
Sure. As I see, it would be nice to fix #92 first. I guess it sould be your task. Then I could try to make the test to work for the new syntax as well as per https://github.com/aaugustin/websockets/pull/90#issuecomment-165035693 which seems a more mechanical work which need a less undersanding of the library/protocol than the previous fix. I would keep those changes in this PR. Plus, if I will have time before you fix #92, I will try to write a server side context manager.

aaugustin:
I tried to minimize your proof of concept and I got to this:

    try:
    
        class Connect:
            """
            This class wraps :func:`connect` on Python 3.5 and above.
    
            It can be used as an asynchronous context manager.
    
            """
            def __init__(self, *args, **kwargs):
                self.connect_coroutine = connect(*self.args, **self.kwargs)
    
            def __await__(self):
                # TODO Check if we can use await here or if we need yield from.
                # Ádam used yield from but I'm pretty sure we can use await.
                return (await self.connect_coroutine)
    
            async def __aenter__(self):
                self.websocket = await self
                return self.websocket
    
            async def __aexit__(self, exc_type, exc_value, traceback):
                await self.websocket.close()
    
            # TODO Check whether we actually need this. PEP 492 implies that we
            # don't. If we do, we need a test case for it.
            __iter__ = __await__
    
    except SyntaxError:
        pass
    
    else:
        connect = Connect
        # Copy over docstring to support building documentation on Python 3.5.
        connect.__doc__ = connect.__doc__

aaugustin:
A few changes explained:

- I removed the `functools.wraps` in favor of explicitly copying the docstring, which is the only thing we need. I think users can easily figure out what's going on if `import connect` gives them a `Connect` class by looking at the source.

- I defined `__iter__` in terms of `__await__` because I always define the "old" way in terms if the "new" way. That leads to cleaner code because the "new" way is superior (at least it's supposed to be...)

- I'm not using _-prefixing for indicating pseudo-private members in this library. Both ways have pros and cons; let's just stay consistent with the current coding style.

- I minimized the number of lines of code.

aaugustin:
If you're unsure about how to write tests, I can take it from there. You've done the hard part :-)

SzieberthAdam:
* `return (await self.connect_coroutine)`: `__await__()` can't contain `await` (https://www.python.org/dev/peps/pep-0492/#list-of-functions-and-methods), it raises a SyntaxError. I checked it before and again.
* self.connect_coroutine = connect(*args, **kwargs): Causes an infinite loop. This is why the original function has to get wrapped in the class. I would stick to `__wrapped__` to match the builtin way.
* Hmm, this is weird: now everything inside one module raises `SyntaxError` in the websockets `__init__` in Python 3.4.3. I was sure I tested it before and it worked. Maybe I missed something. It seems we must go back to the separate module import approach.
* Without `__iter__ = __await__`, those who prefer to use the `yield from` syntax even in version 3.5 will get a `TypeError: 'Connect' object is not iterable` exception. I tested it before and now again.

Next time I make a more detailed comment about what I done and for what reason. I refresh the PO immediately with the new code.

SzieberthAdam:
cff8d12 is tested for Python versions 3.4.3 and 3.5.1. All current tests succeeds but those has to get enhanced. Both `client.py` and `client35.py` examples work on 3.5.1.

I also started to work on the server side but I can put that aside for working on the tests.

SzieberthAdam:
I am busy today so feel free to take it. :) What I suggest that test `await`, `yield from`, and `async with`+`await` syntax for version 3.5 and above.

aaugustin:
Thanks a bunch for double checking the answers to my questions!

I have strong opinions about how to write tests and documentation :-) If that's fine for you, I'll squash your commits and add tests and docs in a new PR.

At this point the most useful contribution you could make is to come up with a proof-of-concept for starting and stopping a server with a context manager.

There's no hurry. I'd like to release 3.0 by the end of the year if possible, but no one's waiting for it either!

aaugustin:
We're on the bleeding edge... https://bitbucket.org/ned/coveragepy/issues/434/indexerror-in-python-35

aaugustin:
Squashed and integrated into #94.

SzieberthAdam:
"At this point the most useful contribution you could make is to come up with a proof-of-concept for starting and stopping a server with a context manager."

I started that yesterday and had some progress. Maybe I can continue it tomorrow.

PR #91, opened 2015-12-15
Update docs to async / await syntax.

### LINKED ISSUE #88: Should examples be switched to Python 3.5 async syntax?
At some point, we should change the examples from:

    @asyncio.coroutine
    def hello():
        websocket = yield from websockets.connect('ws://localhost:8765/')
        # ...

to:

    async def hello():
        websocket = await websockets.connect('ws://localhost:8765/')
        # ...

Pro:

- slightly more concise and readable
- maximizes Python hipster factor

Con:

- doesn't work on Python 3.3 and 3.4 which might generate support questions

Of course, we could point users of Python 3.4 to older versions of the docs.

An alternative would be to show both versions, but I'm afraid that might create confusion.

I'm only raising this question for the docs. I don't think there's any value in considering this change for the library's source until we drop support for Python 3.4. Per the current rules, its end of life should be in March 2019, and we aren't planning that far ahead.

aaugustin:
Fix #88.

### DIFF
--- a/docs/changelog.rst
+++ b/docs/changelog.rst
@@ -17,14 +17,14 @@ Changelog
     ``None`` when the connection was closed. This required checking the return
     value of every call::
 
-        message = yield from websocket.recv()
+        message = await websocket.recv()
         if message is None:
             return
 
     Now it raises a :exc:`~websockets.exceptions.ConnectionClosed` exception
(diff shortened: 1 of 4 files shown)


aaugustin:
As I just said in the ticket, let's put this on hold until Pygments has a new release.

PR #94, opened 2015-12-17
Make connect() usable as an asynchronous context manager.

aaugustin:


aaugustin:
@SzieberthAdam Let me know if you have comments on the final result.

It looks like you can easily modify `ClientServerContextManager.test_basic` to run the server as a context manager as well, around the client, when you work on the server side.

SzieberthAdam:
It seems nice and coherent to me.

PR #98, opened 2016-01-19
Add better extension support for WebSocketServerProtocol

Jyrno42:
This allows one to provide extra keyword arguments to the socket handler or do extra authentication/validation during the handshake and bail out if needed.

For example:

```python
class CustomServerProtocol(WebSocketServerProtocol):
    """ WebSocketServerProtocol that also verifies the current user
         by retrieving and validating `token` value from query parameters.
    """
    def get_handler_kwargs(self, path, get_header):
        query_dict = parse_qs(urlparse(path).query)
        session_key = query_dict.get('token', [''])[0]

        if not is_valid_session(session_key):
            raise websockets.InvalidOrigin("User not authenticated")

        return {
            "session_key": session_key,
        }


@asyncio.coroutine
def client_handler(ws, path, session_key):
    while ws.open:
        # This is only an example
        yield from ws.send({'hello': 'world'})

server = websockets.serve(client_handler, 'localhost', 8001, klass=CustomServerProtocol)
asyncio.get_event_loop().run_until_complete(server)
asyncio.get_event_loop().run_forever()
```

### DIFF
--- a/websockets/server.py
+++ b/websockets/server.py
@@ -59,7 +59,7 @@ class WebSocketServerProtocol(WebSocketCommonProtocol):
         try:
 
             try:
-                path = yield from self.handshake(
+                path, handler_kwargs = yield from self.handshake(
                     origins=self.origins, subprotocols=self.subprotocols,
                     extra_headers=self.extra_headers)
             except Exception as exc:
@@ -75,7 +75,7 @@ class WebSocketServerProtocol(WebSocketCommonProtocol):
                 raise
 
             try:
-                yield from self.ws_handler(self, path)
+                yield from self.ws_handler(self, path, **handler_kwargs)
             except Exception:
                 logger.error("Exception in connection handler", exc_info=True)
                 yield from self.fail_connection(1011)
@@ -143,6 +143,8 @@ class WebSocketServerProtocol(WebSocketCommonProtocol):
                 self.subprotocol = self.select_subprotocol(
                     client_subprotocols, subprotocols)
 
+        handler_kwargs = self.get_handler_kwargs(path, get_header)
+
         headers = []
         set_header = lambda k, v: headers.append((k, v))
         set_header('Server', USER_AGENT)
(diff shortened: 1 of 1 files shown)


Jyrno42:
Note: i also tried running the tests, but got 54 failures with the following exception:

```
OSError: [Errno 98] error while attempting to bind on address ('127.0.0.1', 8642): address already in use
```

aaugustin:
Look at the first test failure.

Everything else just fails because the crashing test prevents the port from being reused.

aaugustin:
It seems to me that this would be more explicit and readable like this:

    def authenticate(path):
        query_dict = parse_qs(urlparse(path).query)
        session_key = query_dict.get('token', [''])[0]
        if is_valid_session(session_key):
            return session_key
        else:
            raise AnyExceptionWillDo()

    @asyncio.coroutine
    def client_handler(ws, path):
        session_key = authenticate(path)
        ...

This version could easily be abstracted into a decorator:

    @asyncio.coroutine
    @require_authentication
    def client_handler(ws, path, session_key):
        ...

It's unclear to me that your proposal would be a good abstraction, amenable to solving a wide range of problems in an intuitive way. For starters, I have no idea how to document it.

Jyrno42:
Currently raising any exception from the handler will cause my error log to be filled with `Exception in connection handler` messages: 

```python
try:
    yield from self.ws_handler(self, path)
except Exception:
    logger.error("Exception in connection handler", exc_info=True)
    yield from self.fail_connection(1011)
    raise
```

So to use your solution i would need to be able to bypass the error log somehow. E.g. by raising a specific exception.

aaugustin:
You should be able to write a decorate that simply doesn't call the handler if the user isn't authenticated.

aaugustin:
I'm still unconvinced that the changes in this PR create a good API, so I'm going to close it, but I'm open to discussing the use case further and making sure there's a decent solution.

Issue #102, opened 2016-02-26
Close / cleanup protocol.worker in the ServerProtocol


dzen:
Hello Aymeric,

We've been seeing lately in our logs a huge amount of:

```
task: <Task pending coro=<WebSocketCommonProtocol.run() running at /home/benoit/Projects/blue/webchouquette/venv34/lib/python3.4/site-packages/websockets/protocol.py:392> wait_for=<Future pending cb=[Task._wakeup()] created at /home/benoit/Projects/blue/webchouquette/venv34/lib/python3.4/site-packages/websockets/protocol.py:123> created at /home/benoit/Projects/blue/webchouquette/venv34/lib/python3.4/site-packages/websockets/protocol.py:615>
```

I think this log is generated each time one of our network/load balancer stuff tries to do a healthcheck. This healthcheck is just a TCP check.

I tried to patch the code( https://github.com/aaugustin/websockets/blob/master/websockets/protocol.py#L642 ) in my machine using:

```python
    if self.worker and not self.worker.done():
        self.worker.cancel()
```

What do you think ?

dzen:
note: I started a server, and run the following code multiple times to simulate our LB:
```python
import socket
s = socket.socket(socket.AF_INET, socket.SOCK_STREAM)
s.connect(('localhost', 8765))
s.send("GET / HTTP/1.1")
s.close()
```

aaugustin:
Get a better load balancer ;-)

Seriously, I always focused on what happens after the handshake completes. I didn't put the same effort on handling connection drops before or while the handshake happens. There's certainly room for improvement but I'm not sure `connection_close()` is the right place.

Can you try inserting at line 634:

    if not self.opening_handshake.done():
        self.opening_handshake.set_result(False)

I'm proposing this because I think the `worker` task is waiting at https://github.com/aaugustin/websockets/blob/master/websockets/protocol.py#L392.

dzen:
I think your fix is better since it unblocks the Task.

PR #104, opened 2016-03-24
Tackle issue #102

RemiCardona:
It took some time to figure out how to do the tests correctly, but it works fine here.


aaugustin:
This looks pretty good, thanks! I'll take a closer look at the test and merge, probably during the week-end.

aaugustin:
I merged the PR, then figured out a simpler way to write the test. Thanks your your contribution!

PR #109, opened 2016-04-25
MANIFEST: add README.rst as needed in setup.py

dzen:
I think README.rst is not present in the tarball. Then setup.py fails :/


aaugustin:
This should be fixed with 3a1174a, which made more drastic changes. Thanks.

If you test locally, make sure you remove the README symlink first.

PR #120, opened 2016-06-06
add example for "ssl usage" and "sharing port usage"

ghostplant:
Hi, I found another way not to modify system library and implement port sharing of HTTP Request and WebSocket Request, which required a deep re-implementation of handler.

This code is also an example of websocket using SSL.

Ghostplant cuiwei13@pku.edu.cn

### DIFF
--- a/example/ssl_server.py
+++ b/example/ssl_server.py
@@ -1,0 +1,65 @@
+#!/usr/bin/env python
+
+import ssl, os
+import asyncio
+import websockets
+
+async def counter(websocket, path):
+    count = 0
+    while True:
+        await asyncio.sleep(0.5)
+        count += 1
+        await websocket.send(str(count))
+
+class HTTPWebSocketServerProtocol(websockets.WebSocketServerProtocol):
+    
+    @asyncio.coroutine
+    def handler(self):
+        try:
+            method, path, version = (yield from self.reader.readline())[:-2].decode().split()
+            if method != 'GET' or not path.startswith('/') or version != 'HTTP/1.1':
+                raise
+            headers = {}
+            while True:
+                line = (yield from self.reader.readline())[:-2]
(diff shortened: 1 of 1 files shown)


ghostplant:
Any comment?

aaugustin:
I'm unsure about this approach. I'm building a startup and moving to a new place with my family, so unpaid OSS work doesn't rank very high on my priorities right now.

I don't want to ship an example that includes a HTTP parser. That stuff belongs to a library, not a code sample.

I'm planning to assess your proposal when I find time, likely in July.

ghostplant:
Okay. This is the only way to share ports without changing existing libraries. After all, websocket is mainly used by JS from an HTML client. If they don't share ports, it's easier for users to get into trouble on configurating multiple paisr of PEM certificates.
Just like issues/111, the current example "showtime.html" using SSL mode doesn't work in firefox, and "Malformed HTTP message" will be reported. After sharing port in the added example, this problem gets fixed.

aaugustin:
I looked at this again today.

I appreciate that you took the time to submit this example as a pull request.

Unfortunately it's very far from how I would approach these problems and what I consider best practices. I wouldn't put this in production and I'm not comfortable recommending this approach, the main reasons being:

- there's a bunch of security vulnerabilities in the HTTP parsing (look at the countermeasures in websockets/http.py for some examples)
- I don't think port sharing is a good idea in general, or even useful, and if you want to do it I'd rather you use Twisted than websockets
- I don't want to teach people about TLS that way.

ghostplant:
@aaugustin Thanks, it doesn't matter since I already dropped dependency with websockets a year of because of some other bugs in this module (easily crash under huge quantities of connection, etc). Currently I found that python3-tornado seems to be a better choice since it is a lot more faster.

PR #127, opened 2016-09-02
server: don't print exceptions for network errors during handshake

### LINKED ISSUE #126: harsh warnings on reset connections during handshake
At work, our websocket daemon is behind an HAproxy. Said reverse proxy is configured to probe our daemon to see if it's alive. Since 181acf3e29499ecff61b5e6892d73bae5332292e, this health check generates a _lot_ of logs which look like the following:

``` python-traceback
WARNING 2016-09-02 09:42:19,976 websockets.server Error in opening handshake
  File "/home/rcardona/Projects/websockets/websockets/server.py", line 64, in handler
    extra_headers=self.extra_headers)
  File "/usr/lib/python3.4/asyncio/coroutines.py", line 98, in throw
    return self.gen.throw(exc)
  File "/home/rcardona/Projects/websockets/websockets/server.py", line 143, in handshake
    path, headers = yield from read_request(self.reader)
  File "/usr/lib/python3.4/asyncio/coroutines.py", line 98, in throw
    return self.gen.throw(exc)
  File "/home/rcardona/Projects/websockets/websockets/http.py", line 43, in read_request
    request_line, headers = yield from read_message(stream)
  File "/usr/lib/python3.4/asyncio/coroutines.py", line 98, in throw
    return self.gen.throw(exc)
  File "/home/rcardona/Projects/websockets/websockets/http.py", line 83, in read_message
    start_line = yield from read_line(stream)
  File "/usr/lib/python3.4/asyncio/coroutines.py", line 98, in throw
    return self.gen.throw(exc)
  File "/home/rcardona/Projects/websockets/websockets/http.py", line 103, in read_line
    line = yield from stream.readline()
  File "/usr/lib/python3.4/asyncio/coroutines.py", line 98, in throw
    return self.gen.throw(exc)
  File "/usr/lib/python3.4/asyncio/streams.py", line 418, in readline
    yield from self._waiter
  File "/usr/lib/python3.4/asyncio/futures.py", line 388, in __iter__
    yield self  # This tells Task to wait for completion.
  File "/usr/lib/python3.4/asyncio/tasks.py", line 286, in _wakeup
    value = future.result()
  File "/usr/lib/python3.4/asyncio/futures.py", line 277, in result
    raise self._exception
  File "/usr/lib/python3.4/asyncio/selector_events.py", line 582, in _read_ready
    data = self._sock.recv(self.max_size)
ConnectionResetError: [Errno 104] Connection reset by peer
```

This is due to how HAproxy implements its healthcheck (and I'm told it's quite a standard practice): it opens a connection and resets it immediately (instead of properly closing it, as one would expect). This is to prevent leaving sockets in the `FIN_WAIT2` or (most likely) `TIME_WAIT` states on proxy side.

To reproduce the issue, `nmap -p <port> <hostname>` from a _different_ host does the trick (if done locally, the connection is closed and no exception is raised).

As far as I'm concerned, just ignoring this kind of exception at this point in time (the handshake) does the trick for me. In any case, a connection reset is not an actual failure that `websockets` could have prevented in any meaningful way. It's just one of the many things that may happen when dealing with sockets.

My proposed solution is to simply log `ConnectionError` exceptions (which include `ConnectionResetError`) as `INFO` without a traceback.

RemiCardona:
Since 181acf3, full tracebacks are printed if a connection is reset
during handshake.  Connection resets are to be expected, even in
controlled environments.  Simply log them as info rather than warning,
and don't print a full traceback.

Closes #126.

### DIFF
--- a/websockets/server.py
+++ b/websockets/server.py
@@ -62,6 +62,9 @@ class WebSocketServerProtocol(WebSocketCommonProtocol):
                 path = yield from self.handshake(
                     origins=self.origins, subprotocols=self.subprotocols,
                     extra_headers=self.extra_headers)
+            except ConnectionError as exc:
+                logger.info('Connection error during opening handshake', exc_info=True)
+                raise
             except Exception as exc:
                 if self._is_server_shutting_down(exc):
                     response = ('HTTP/1.1 503 Service Unavailable\r\n\r\n'
@@ -89,6 +92,11 @@ class WebSocketServerProtocol(WebSocketCommonProtocol):
 
             try:
                 yield from self.close()
+            except ConnectionError as exc:
+                if self._is_server_shutting_down(exc):
+                    pass
+                logger.info('Connection error in closing handshake', exc_info=True)
+                raise
             except Exception as exc:
                 if self._is_server_shutting_down(exc):
                     pass


JCharante:
I would love it if this was merged.

aaugustin:
I'm OK with merging this (and Remi has commit access).

I believe test coverage is at 100% and I'd like it remain there, but that isn't an absolute prerequisite for error handling paths.

PR #128, opened 2016-09-11
Simplify combined consumer-producer example.

cjerdonek:
In the docs [here](https://websockets.readthedocs.io/en/stable/intro.html#both), where you provide an example combining the consumer and producer examples, you say:

> This code looks convoluted. If you know a more straightforward solution, please let me know about it!

This patch provides what seems like the more natural approach. The idea is to use  [`asyncio.gather()`](https://docs.python.org/3/library/asyncio-task.html#asyncio.gather) instead of [`asyncio.wait()`](https://docs.python.org/3/library/asyncio-task.html#asyncio.wait).

I confirmed that this approach works locally.

### DIFF
--- a/docs/intro.rst
+++ b/docs/intro.rst
@@ -59,7 +59,7 @@ Consumer
 
 For receiving messages and passing them to a ``consumer`` coroutine::
 
-    async def handler(websocket, path):
+    async def receiver_handler(websocket, path):
         while True:
             message = await websocket.recv()
             await consumer(message)
@@ -73,7 +73,7 @@ Producer
 
 For getting messages from a ``producer`` coroutine and sending them::
 
-    async def handler(websocket, path):
+    async def producer_handler(websocket, path):
         while True:
             message = await producer()
             await websocket.send(message)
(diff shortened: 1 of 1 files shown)


aaugustin:
Essentially you're running the producer and consumer coroutines. Indeed, this looks easier than forcing them in the same `while` loop.

```python
        while True:
            await gathered
```

isn't needed as the coroutines don't return.

Actually, with this approach, all you need is:

```python
    async def handler(websocket, path):
        # This call prevents the handler from exiting, which would
        # terminate the websocket connection.
        await asyncio.wait([
            receiver_handler(websocket, path),
            producer_handler(websocket, path),
        ])
```

aaugustin:
I'd like to keep both the "two parallel loops" and "one complicated loop" approaches in the docs, in this order, because this is the sort of thing that needs to click when learning asyncio.

cjerdonek:
> Actually, with this approach, all you need is:

Great!

Should I modify my PR as you described or let you take care of it?

cjerdonek:
I've updated the PR per your suggestions.

aaugustin:
Chris, I'm sorry, but I lost track of this PR and a similar change got committed independently :-(

I preferred a sentence from your version and committed it in 4a57039.

cjerdonek:
@aaugustin No problem, thanks!

PR #132, opened 2016-09-28
Make select_subprotocol a static method

lgrahl:
As I need to use this function from the outside (feel free to ask why if you want to) and the method can be static, I've made it static. Nothing special. :)

### DIFF
--- a/websockets/server.py
+++ b/websockets/server.py
@@ -195,7 +195,8 @@ class WebSocketServerProtocol(WebSocketCommonProtocol):
 
         return path
 
-    def select_subprotocol(self, client_protos, server_protos):
+    @staticmethod
+    def select_subprotocol(client_protos, server_protos):
         """
         Pick a subprotocol among those offered by the client.
 


aaugustin:
Wouldn't this prevent users from subclassing the protocol and implementing a select_protocol that depends on instance variables? I don't know if you can replace a static method by a regular method in a subclass.

lgrahl:
Good question! That shouldn't be a problem, it's even possible to call the inherited method via super:

```python
class A:
    @staticmethod
    def a(arg):
        print(A, arg)

class B(A):
    def a(self, arg):
        super().a(arg)
        print(self, arg)

a = A()
a.a('hello')
b = B()
b.a('hello')
```

lgrahl:
We could also use `@classmethod` instead of `@staticmethod`.

aaugustin:
FTR this change broke the tests on some Python versions because of https://bugs.python.org/issue23078.

PR #134, opened 2016-10-22
Update intro.rst with a better producer/consumer

wojtossfm:
See gist for example test cases https://gist.github.com/wojtossfm/aede13446bb3b9e12942109f505058c1

The example code for the docs is prone to a race condition of the state of the tasks. If people replicate it they might wonder why they are losing messages (due to the cancel calls). The proposed change avoids the race condition. I tried to change as little as possible in the example.

If I'm missing something or something is unclear then let me know.

Just thought I'd chip in if I can. Great library by the way.

### DIFF
--- a/docs/intro.rst
+++ b/docs/intro.rst
@@ -111,7 +111,7 @@ messages on the same connection.
                 producer_task = asyncio.ensure_future(producer())
                 pending.add(producer_task)
 
-(This code looks convoluted. If you know a more straightforward solution, 
+(This code looks convoluted. If you know a more straightforward solution,
 please let me know about it!)
 
 Registration
--- a/docs/intro.rst
+++ b/docs/intro.rst
@@ -91,27 +91,26 @@ messages on the same connection.
 ::
 
     async def handler(websocket, path):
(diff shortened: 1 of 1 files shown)


aaugustin:
Indeed it's a good idea to avoid losing messages.

It looks like you're never removing completed tasks from the `pending` set. Are you sure this code works?

I'd expect something like:
```python



The code is even more convoluted that it used to be, I think the apology still applies, can you leave it there?

wojtossfm:
But you don't need to remove tasks from the pending set. asyncio wait returns two sets - completed tasks and pending tasks. If a task is completed it is no longer in the pending set. Have a look at the gist in my comment. Go ahead and run it in case. It has similar code for asyncio (just without websockets) as examples what was wrong with the previous example and how this one works.

aaugustin:
Right! I had missed that `done, pending = await asyncio.wait(...)` mutates `pending`. Sorry.

I'll do some testing when I have a bit more time, but at first sight, it looks good. Thanks!

aaugustin:
The docs now contain a much more simple example, making this discussion moot.

Thanks for your help anyway and sorry for not being responsive!

PR #144, opened 2016-11-23
Improve getting started for -both- case

bajamircea:
Improve example in getting started for handling both listening and a producer


aaugustin:
Thanks!

PR #153, opened 2017-01-11
Added client and server showing how to send/receive requests

RojavaCrypto:
This is an example showing how to build wrapper classes for websockets that lets you run code like:

```
async def hello():
    async with Client('ws://localhost:8765') as client:
        error, result = await client.query("myapp_do_something",
                                           12, "foo")
        print(error, result)
```

It works by using a dict of {int id: future}. Server responses are used to lookup a future through the integer id, and route the response to the relevant running task (asyncio thread).

### DIFF
--- a/example/client_reqid.py
+++ b/example/client_reqid.py
@@ -1,0 +1,88 @@
+#!/usr/bin/env python
+
+import asyncio
+import json
+import random
+import websockets
+
+def create_random_id():
+    max_uint32 = 2**32 - 1
+    return random.randint(0, max_uint32)
+
+# This example shows how to multiplex requests from a server
+# using asyncio Future.
+
+class Client:
+
+    def __init__(self, websockets_path):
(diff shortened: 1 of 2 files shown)


aaugustin:
Separating calls to `__aenter__` and `__aexit__` like this isn't a pattern I'm going to encourage.

RojavaCrypto:
Well what else do you suggest?

aaugustin:
Use the standard APIs:

In your `__aenter__`: `self.websocket = yield from websockets.connect(...)`
http://websockets.readthedocs.io/en/stable/api.html#websockets.client.connect

In your `__aclose__`: `yield from self.websocket.close()`
http://websockets.readthedocs.io/en/stable/api.html#websockets.protocol.WebSocketCommonProtocol.close

(add exception handling etc. as desired)

The context manager avoids a try/finally in some use cases, but your use case isn't one of these.

PR #171, opened 2017-04-06
Excplicitly use utf-8 to read README file

bachp:
We already know it is UTF-8 so we can make it explicit.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -7,7 +7,7 @@ root_dir = os.path.abspath(os.path.dirname(__file__))
 
 description = "An implementation of the WebSocket Protocol (RFC 6455)"
 
-with open(os.path.join(root_dir, 'README.rst')) as f:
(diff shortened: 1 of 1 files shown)


aaugustin:
Does this fix a bug? Or is it a personal preference?

If it's a personal preference, I can only conclude that our preferences diverge ;-)

bachp:
I actually hit this when trying to package this for Yocto (see [here](https://github.com/bachp/meta-homeassistant/blob/master-next/recipes-devtools/python/python3-websockets_3.3.bb)). The issue seems that there are some systems where UTF-8   is not the default encoding 😢, and python was throwing a ascii decode error.

So it's not only personal taste 😄

aaugustin:
Thanks for bringing this to my attention. There were two open() calls in setup.py; I fixed them both.

PR #172, opened 2017-04-11
Only log `failing the WebSocket connection` when code is not 1006

### LINKED ISSUE #167: logs overwhelmed by Failing the WebSocket connection: 1006
Seeing this a lot in logs when closing a websocket connection.

Code `1006` seems caused by `readexactly` hitting `EOF`, which seems fine when closing websocket connection.

messense:
Fixes #167

### DIFF
--- a/websockets/protocol.py
+++ b/websockets/protocol.py
@@ -615,13 +615,13 @@ class WebSocketCommonProtocol(asyncio.StreamReaderProtocol):
     @asyncio.coroutine
     def fail_connection(self, code=1011, reason=''):
         # 7.1.7. Fail the WebSocket Connection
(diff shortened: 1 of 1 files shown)


RemiCardona:
Again, could you paste/gist a full backtrace please? That would be helpful to evaluate your PR.

Thanks

messense:
It's fairly simple, no exception raised, just some logging:

```
2017-03-29 11:33:57,429 INFO websockets.protocol.fail_connection:613   Failing the WebSocket connection: 1006
```

aaugustin:
I think we should downgrade some logging calls from INFO to DEBUG.

@RemiCardona I think that was one of your early concerns with this library. If you'd like to review the logging, make it more consistent and improve levels, be my guest ;-)

PR #175, opened 2017-04-12
Speedup websocket mask using C copied from Tornado codebase

messense:
Is there anyway to do a benchmark?

Is there a license issue with the Tornado codes?


RemiCardona:
Personally, I'd be _extremely_ weary of adding C extensions to this project without:
* fully measuring the impact using benchmarks
* making sure the extension stays optional (much like asyncio's own C extensions in 3.6) and that tests are adapted to exercise **all** code paths.

That being said, I did do a few micro benchmarks around the masking operation.
```python
from itertools import cycle

data = b'abcdef' * 10000
mask_bits = b'L\xde\xcb:'  # obtained through struct.pack('!I', random.getrandbits(32))

# original version
masked_data = bytes(b ^ mask_bits[i % 4] for i, b in enumerate(data))

# new version
masked_data2 = bytes(b ^ m for b, m in zip(data, cycle(mask_bits)))
```

Using ipython's `%timeit`, here are some results:
* 3.4: 10.7ms to 6.39ms → 40% speedup
* 3.6: 9.65ms to 5.91ms → 38% speedup

However, with `data = b'abcdef'`, timeit reports 

> The slowest run took 4.65 times longer than the fastest. This could mean that an intermediate result is being cached.

whereas the original version looks much more stable.

So this looks promising, but I have no idea what the impact of this is on the overall performance of websockets. @messense the same applies to your PR, we need to be able to gauge whether adding a C extension to a pure-python project is worth the hassle.

Cheers

RemiCardona:
Addendum: with really short data, both versions are equally unstable on 3.6, so 🤷‍♀️

messense:
with `data = b'abcdef'`

```python
In [22]: %timeit pure_python_mask1(mask_bits, data)
100000 loops, best of 3: 2.22 µs per loop

In [23]: %timeit pure_python_mask2(mask_bits, data)
100000 loops, best of 3: 2.1 µs per loop

In [24]: %timeit websocket_mask(mask_bits, data)
The slowest run took 8.93 times longer than the fastest. This could mean that an intermediate result is being cached.
1000000 loops, best of 3: 235 ns per loop
```

setup.py
```diff
@@ -1,7 +1,71 @@
...
+version of Python.
+Debian and Ubuntu users should issue the following command:
+    $ sudo apt-get install build-essential python-dev
+RedHat and CentOS users should issue the following command:
+    $ sudo yum install gcc python-devel
+Fedora users should issue the following command:
+    $ sudo dnf install gcc python-devel
+If you are seeing this message on OSX please read the documentation
+here:
+http://api.mongodb.org/python/current/installation.html#osx
```
aaugustin: I really, really don't want to handle support questions about MongoDB's documentation.

websockets/speedups.c
```diff
@@ -0,0 +1,68 @@
+/*
+Copyright 2009 Facebook
```
aaugustin: I really don't want to take responsibility for shipping code copyrighted by Facebook either.

They have lawyers and money. I don't.

websockets/speedups.c
```diff
@@ -0,0 +1,68 @@
...
+
+    return result;
+}
+
+static PyMethodDef methods[] = {
+    {"websocket_mask",  websocket_mask, METH_VARARGS, ""},
+    {NULL, NULL, 0, NULL}
+};
+
+#if PY_MAJOR_VERSION >= 3
```
aaugustin: I don't see the point of this conditional given that websockets only supports Python 3.

aaugustin:
Thanks for the pull request.

I'm in favor of shipping a C extension to accelerate low-level operations.

One way to figure out the effect is to run the Autohahn benchmark. Instructions for running it are included in the `compliance` directory.

I'm uncomfortable all the pieces that appear to be copy-pasted from other projects and that don't necessarily serve a purpose here. `setup.py` must remain trivial.

A prerequisite for merging this PR is to figure out how to measure the coverage of C code, like Python code. That would have caught the useless conditional branch for Python 3 vs Python 2 that I saw when I skimmed the diff.

There must be a way to run tests with the speedups and without the speedups. The combination of the two runs must have 100% branch coverage of the C and the Python code. This [toy project](https://github.com/aaugustin/sudoku) which I wrote to learn about C extension can provide some hints in this area.

The C module itself is fairly short. I'll read it when I have more time. I'll have to check that it doesn't leak memory (in the absence or presence of errors).

websockets/speedups.c
messense: Tornado is licensed under [Apache 2.0](https://github.com/tornadoweb/tornado/blob/master/LICENSE) license.

If you are ok with it, I'll address the issues raised in review.

aaugustin:
Thanks messense for getting the ball rolling on this. I'm a bit more optimistic about #211 which doesn't contain code copyrighted by Facebook, under a license that isn't fully compatible with websockets's license.

I'm closing in favor of #211, but comparing the two PRs and suggesting improvements in #211 based on ideas found in this one.

aaugustin:
@messense - I didn't end up merging your version, but I'm making you a collaborator to this repo to recognize your contribution there!

Issue #93, opened 2015-12-16
Add support for extensions and implement RFC 7692


aaugustin:
Until August 2015, I could get away with saying that no extensions were registered. Now there's one.

aaugustin:
If you're interested in this feature, upvote this issue. I won't implement it until someone needs it.

RouquinBlanc:
Hi Aaugustin,

Trying to move away from our nodejs websocket server, we will run into an issue using your implementation: the lack of support for websocket compression (permessage-deflate), which is now quite deployed on browsers and in nodejs, and saves a lot of bandwidth for our mobile clients. Do you have any plan concerning this one?

Thanks
Jonathan

aaugustin:
Hello, while I'd like websockets to gain this feature, I'm unlikely to work on it in my free time for the next two months.

If you'd like to discuss sponsoring this feature, please email me: aymeric.augustin@fractalideas.com. Thanks!

aaugustin:
I started a branch here: https://github.com/aaugustin/websockets/tree/extensions

aaugustin:
Pull request for that branch: https://github.com/aaugustin/websockets/pull/178

PR #178, opened 2017-05-03
Add support for extensions (#93)

aaugustin:


websockets/framing.py
```diff
@@ -182,9 +188,9 @@ def check_frame(frame):
     if frame.rsv1 or frame.rsv2 or frame.rsv3:
         raise WebSocketProtocolError("Reserved bits must be 0")
 
-    if frame.opcode in (OP_CONT, OP_TEXT, OP_BINARY):
+    if frame.opcode in [OP_CONT, OP_TEXT, OP_BINARY]:
```
RemiCardona: What about a `set` here (and below) instead? As far as perf goes, tuple creation is somewhat faster than that of lists so I would suggest keeping the original.

websockets/framing.py
```diff
@@ -95,13 +95,14 @@ def read_frame(reader, mask, *, max_size=None, extensions=()):
...
-    rsv3 = bool(head1 & 0b00010000)
+    # While not very Pythonic, this is marginally faster than calling bool().
+    fin = True if head1 & 0b10000000 else False
+    rsv1 = True if head1 & 0b01000000 else False
+    rsv2 = True if head1 & 0b00100000 else False
+    rsv3 = True if head1 & 0b00010000 else False
     opcode = head1 & 0b00001111
 
-    if bool(head2 & 0b10000000) != mask:
+    if (True if head2 & 0b10000000 else False) != mask:
```
RemiCardona: Why not go with
```python
if head2 & 0b10000000 != mask:
```
instead? I feel like I'm missing something :)

aaugustin:
It's work in progress, I want to implement the compression extension and pass the AutoBahn test suite before merging this PR.

websockets/framing.py
aaugustin: There's a comment a few lines above that says:

```# While not Pythonic, this is marginally faster than calling bool().```

I think I'd stick with the more readable version these days; my opinions on micro-optimizations changed.

File: websockets/framing.py (hunk 1 above)
Comment (aaugustin) [inline websockets/framing.py]: I didn't know the performance difference.

It would be faster to create the tuple or list at module level anyway...

**Comment by RouquinBlanc** [comment]
Hi Augustin,

I may have an idea of why all those tests are still failing on autobahn.

First of all, thoses tests all failed because the server closed the connection too early; looking at the reason, it is 1007, which leads to UnicodeDecodeError.

Under debugger, I managed to reach the point of failure, and it happens that, because of the fragmentation, the current is cut right in the middle of an utf-8 character, which causes the issue.

I don't think you can decode 1 frame at a time, but should instead wait for the last fragment to decode. But I'm wondering why this issue is not appearing as well without compression? Could it be a latent problem?

And... wow! all the other tests are PASS! This is looking pretty encouraging 👍

**Comment by RouquinBlanc** [comment]
Humm... refining of previous statement: now I see the problem here: (protocol.py:466
```
        # Shortcut for the common case - no fragmentation
        if frame.fin:
            return frame.data.decode('utf-8') if text else frame.data <<<< raising UnicodeDecodeError
```
... that means that we received a 1 frame message ending with a non-utf8 character. Here is the data:
```
b'<b>Case Description</b><br/><br/>
[... cut, was 4096 long ...]
<br>MESSAGE FRAGMENT 1:<br>Hello-\xc2'
```
it's probably perfectly legal in python2. I need to look at what the spec says about TEXT messages, are they supposed to be utf-8 decodable?

**Comment by RouquinBlanc** [comment]
Have the answer from RFC 6455:

Section 5.6
```
Text

      The "Payload data" is text data encoded as UTF-8.  Note that a
      particular text frame might include a partial UTF-8 sequence;
      however, the whole message MUST contain valid UTF-8.  Invalid
      UTF-8 in reassembled messages is handled as described in
      Section 8.1.
```
And Section 8.1
```
   When an endpoint is to interpret a byte stream as UTF-8 but finds
   that the byte stream is not, in fact, a valid UTF-8 stream, that
   endpoint MUST _Fail the WebSocket Connection_.  This rule applies
   both during the opening handshake and during subsequent data
   exchange.
```

So that would mean that (correct me if I'm wrong):

- It is perfectly right to fail the connection if you receive a message (not frame) which is not UTF-8
- But it is not correct to attempt decoding frames of a fragmented message individually

**Comment by aaugustin** [comment]
The code path you're referring to only kicks in when there's no fragmentation -- that is when the first data frame has the FIN flag set.

Just below there's a loop that feeds the data from successive frames to an incremental UTF-8 decoder and reassembles fragmented messages.

Perhaps I messed up the encoding/decoding code and I'm setting the FIN flag when it shouldn't be set, thus ending up incorrectly in the "no fragmentation" code path.

**Comment by RouquinBlanc** [comment]
Some of the tests fails on not fragmented code path, some on the fragmented one.

**Comment by RouquinBlanc** [comment]
Look specifically at test 12.4.5.
```
Case Description

Send 1000 compressed messages each of payload size 4096, auto-fragment to 0 octets. Use default permessage-deflate offer.
```

Looking at the test definition in autobahntestsuite/case/case12_x_x.py, they don't validate that what is sent as a message is utf-8 compliant.

Adding a simple `msg.encode('utf-8')` there before sending to the server under test raises an exception on the 35th message
```
  File "/Users/homerunisgood/virtualenvs/venv27/lib/python2.7/site-packages/autobahntestsuite/case/case12_x_x.py", line 160, in sendOne
    msg.encode('utf-8')
exceptions.UnicodeDecodeError: 'ascii' codec can't decode byte 0xc2 in position 1976: ordinal not in range(128)
```
How can autobahn pass this test?

**Comment by RouquinBlanc** [comment]
Adding a check in autobahn case_12_x_x.py to ensure it sends only messages which are utf-8 compliant, all the tests are pass

**Comment by aaugustin** [comment]
Thanks for your investigation, that's super helpful.

I inspected the code of the test suite and I read it like you.

`self.testData = open(fn, 'rb').read()` => returns an utf-8 encoded bytestring

`msg = self.testData[idxFrom:idxTo]` => may cut in the middle of a multibyte utf-8 character

Now I'm surprised that it contains such a bug, I assume at least some other implementations would crash like mine?

**Comment by aaugustin** [comment]
So, yeah, we aren't the first ones to notice, this is https://github.com/crossbario/autobahn-testsuite/issues/71.

**Comment by aaugustin** [comment]
There's a significant gap in the current design of this PR: the API for extensions doesn't handle negotiation.

We need two level of objects:

- an "extension factory" that handles negotiation: 1 for a server, 1 for a client
- an "extension instance" that processes frames: 1 per connection for a server, 1 for a client (for the single connection of that client)

For consistency and clarity, it seems best to separate these two concepts on the client side as well as on the server side, even if the factory will only produce one instance.

Naming them will be fun.

This is completely missing from the current design of the API. Furthermore, the factories will look quite different between the client and the server side, while the instances should be mostly identical, given that the protocol is symmetrical.

----

The negotiation phase looks like this.

On the client side:

1. an extension factory is configured with some parameters (with sane defaults)
2. the resulting parameters are sent to the server in the handshake request
3. the server sends something (or nothing) back in the handshake response
4. the extension may accept the server's answer and perhaps configure itself accordingly, or reject it and fail the connection

On the server side:

1. an extension factory is configured with some parameters (and sane defaults)
2. the client sends something (or nothing) in the handshake request
3. the extension may accept the client's request and perhaps configure itself accordingly, or reject it and fail the connection, or ignore it and disable itself
4. the resulting parameters are sent to the client in the handshake response

So we need the following APIs:

```python
class ClientExtensionFactory:

    name = ...

    def __init__(self, **kwargs):
        """
        Configure the extension.

        Define negotiation rules. May accept callable arguments for complex logic.
        """

    def get_request_params(self):
        """
        Build request parameters.

        They should derive directly from the configuration.
        """

    def process_response_params(self, params):
        """"
        Process response parameters.

        Return configured extension instance.

        TODO: normalize exception to signal failure
        """
        ...
        return instance  # provides .encode() and .decode()

class ServerExtensionFactory:

    name = ...

    def __init__(self, **kwargs):
        """
        Configure the extension.

        Define negotiation rules. May accept callable arguments for complex logic.
        """

    def process_request_params(self):
        """"
        Process request parameters.

        Return response params and configured extension instance.

        TODO: normalize exception to signal failure
        """
        return (
            response_params,
            instance,  # provides .encode() and .decode()
        )

```

```
class ExtensionInstance:

    def decode(self, frame):
        """Decode an incoming frame."""
        return frame

    def encode(self, frame):
        """Encode an outgoing frame."""
        return frame

```

ExtensionInstance is described as a class here.

It could also be a pair of functions. These functions could still be bound methods on an instance, if carrying state in a class is more convenient. The drawback of the pair of functions is that it's harder to add something to the API, while a class makes it possible to add more methods later if needed (like cleanup methods for example). So it's probably more future proof to go with a class instance.

aaugustin:
The current implementation does provide these two levels, but the first one is extremely limited and barely visible: `WebSocketServerProtocol.process_extensions` has a hardcoded check for `permessage-deflate` that instanciates a `PerMessageDeflate` extension instance (which, in turn, does the bulk of the negotiation).

I believe the API I outlined above is better than the current implementation because it separates negotiation from processing cleanly, and these are separate concerns.

RouquinBlanc:
I like the factory concept. It will make it much easier for future extensions, and seems indeed saner then the current instantiation. You're planning to have compression on by default on both sides, right? Note that it would be great to keep a way to quickly/easily be able to force disable compression at server or client instantiation; one common use case is when you run your websockets with network recording, it is much more convenient to have compression off in order to read the content!

aaugustin:
While I still have a bunch of improvements on my TODO list, with my latest push, I believe the general design is now in good shape. Feedback welcome!

aaugustin:
Of course test coverage is very much lacking. I'll work on it.

RouquinBlanc:
I had a syntax error in the permessage deflate factories, once fixed it runs like a charm.

Quick note: both deflate factories should be exported, in case you want to specify other options than those by default

Will have a closer look

RouquinBlanc:
I had an error in Safari:
```
WebSocket connection to 'ws://server:8080/' failed: Invalid UTF-8 sequence in header value
```
Was due to the following request:
```
GET / HTTP/1.1
Upgrade: websocket
Connection: Upgrade
Host: server:8080
Origin: http://server
Pragma: no-cache
Cache-Control: no-cache
Sec-WebSocket-Key: BXicWBw2TYXXUJnyHsnfGA==
Sec-WebSocket-Version: 13
Sec-WebSocket-Extensions: x-webkit-deflate-frame
User-Agent: Mozilla/5.0 (Macintosh; Intel Mac OS X 10_11_6) AppleWebKit/603.3.8 (KHTML, like Gecko) Version/10.1.2 Safari/603.3.8
```
And the reply:
```
HTTP/1.1 101 Switching protocols
Server: Python/3.4 websockets/3.3
Sec-WebSocket-Extensions: 
Upgrade: WebSocket
Connection: Upgrade
Sec-WebSocket-Accept: g4Hr94fj3OjAoBZpF4yyxqS8+3s=
```
Safari requests `x-webkit-deflate-frame` which we do not support, and does not allow empty headers in the reply.

It's a good use case, I guess we will have to make a round on all navigators to check how they behave!

aaugustin:
Good catch — RFC 6455 says things about null vs. empty values. Generally an empty value represented by the empty string in Python isn't acceptable. The null value represented by `None` in Python corresponds to the lack of the header.

aaugustin:
The code is complete and mostly tested.

If someone feels courageous and want to make a review, that would be really helpful!

I still have to:

- [ ] complete test coverage
- [ ] add documentation
- [ ] add changelog

To be perfectly RFC-compliant -- but this isn't blocking for merging IMO -- I also have to:

- [ ] handle multiple `Sec-Websocket-Headers`
- [ ] add logic to handle properly multiple offers of the `permessage-deflate` extension with different parameters

(plus a few other ideas which require some investigation, I'm keeping them in my own TODO list)

RouquinBlanc:
Tested our solution with compression on Chrome, Firefox and Opera without any problem.
Safari does not implement permessage-deflate (instead using their own thing as discussed).
Internet Explorer does not enable compression by default.

aaugustin:
And test coverage is done! This is adding about 50% to the amount in the project 🤓

aaugustin:
@RouquinBlanc Thanks, that's very helpful!

I never bothered exercising the implementation with a browser (!)

aaugustin:
I'm seeing failures in the autobahn tests: some of 12.4.* and all of 13.7.*.

aaugustin:
12.4.* is a bug in the autobahn test suite.

aaugustin:
13.7 was failing because I didn't handle multiple offers of permessage-deflate. Now I do.

aaugustin:
I think this is done, except for the documentation.

What a marathon.

aaugustin:
Merge now, discuss later!

aaugustin:
> :trollface:

PR #181, opened 2017-05-05
Make serve an async context manager.

### LINKED ISSUE #86: Consider making websocket protocols context managers
From https://github.com/aaugustin/websockets/issues/77#issuecomment-152921813

it would also be nice to use pythons context manager syntax to guarantee client code can be robust without any need for checking on their end.

something like this:

``` python
def echo_handler(websocket):
    with websocket as ws:
        # as long as the socket is open do your thing
        while True:
            message = yield from ws.recv()
            yield from ws.send(message)
```

it's not necessarily pretty (it took 3 indents to write anything useful), but it'll take some responsibility off a client.

---

I'm not sold on this approach yet but I think it's interesting and deserves its own ticket.

aaugustin:
Fix #86.

### DIFF
--- a/docs/changelog.rst
+++ b/docs/changelog.rst
@@ -6,6 +6,9 @@ Changelog
 
 *In development*
 
+* :func:`~websockets.server.serve` can be used as an asynchronous context
+  manager on Python ≥ 3.5.
+
 * Added support rejecting incoming connections by customizing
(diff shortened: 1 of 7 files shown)


aaugustin:
I'd love it if someone could confirm that this seems sensible ;-)

PR #186, opened 2017-05-18
Quick and dirty prototype of compression/decompression

RouquinBlanc:
Hi Augustin,

In order to resurrect my interest on compression support for websockets, I have done a (very) quick and (VERY) dirty implementation of compression/decompression optional support using what you started implementing.

I hope that some things are of interest. I did a simple test communicating with wscat, and it just works 👍

There are many more things to be done. I will try to push a bit further if I find time.

### DIFF
--- a/compliance/fuzzingclient.json
+++ b/compliance/fuzzingclient.json
@@ -6,6 +6,6 @@
    "servers": [{"agent": "websockets", "url": "ws://localhost:8642", "options": {"version": 18}}],
 
    "cases": ["*"],
-   "exclude-cases": [],
+   "exclude-cases": ["12.*", "13.*"],
    "exclude-agent-cases": {}
 }
--- a/compliance/fuzzingserver.json
+++ b/compliance/fuzzingserver.json
@@ -7,6 +7,6 @@
    "webport": 8080,
 
    "cases": ["*"],
-   "exclude-cases": [],
+   "exclude-cases": ["12.*", "13.*"],
    "exclude-agent-cases": {}
 }
--- a/websockets/extensions.py
+++ b/websockets/extensions.py
@@ -1,0 +1,102 @@
+import zlib
+
+from .framing import CTRL_OPCODES, OP_CONT
+
+
+__all__ = ['PerMessageDeflate']
+
+
+_EMPTY_UNCOMPRESSED_BLOCK = b'\x00\x00\xff\xff'
+
+
+class PerMessageDeflate:
+    """
+    Compression Extensions for WebSocket (`RFC 7692`_).
+
+    .. _RFC 7692: http://tools.ietf.org/html/rfc7692
+
+    """
+
+    # This class implements the server-side behavior by default.
+    # To get the client-side behavior, set is_client = True.
+    is_client = False
+
+    def __init__(self):
+        # Currently there's no way to customize these parameters.
+        self.server_no_context_takeover = False
+        self.client_no_context_takeover = False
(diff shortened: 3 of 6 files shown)


RouquinBlanc:
OK after a pass on RFC7692, here are the things which need to change:

1. Written in the extension name: permessage-xxx; the compression/decompression needs to happen at the message level, not frame level. It does not mean that the per frame evaluation of extensions is wrong though, as some extensions may have a per frame impact.
2. The Frame check procedure needs to evolve; with extensions some reserved bits are going to be in use
3. Interesting check to put in place:
> An endpoint MUST NOT set the "Per-Message Compressed" bit of control
   frames and non-first fragments of a data message.  An endpoint
   receiving such a frame MUST _Fail the WebSocket Connection_.
4. the permessage decompression must reset the rsv1 bit to zero before giving the message to next extension

aaugustin:
Argh. I was working on the same feature in my branch, but it wasn't ready for submitting as a PR.

I pushed my work-in-progress state here: https://github.com/aaugustin/websockets/pull/178/commits/aab9747cb2c426d9d023bb9d5de4dcc2efa5dff8

aaugustin:
It looks like we've taken a similar approach.

We both noticed that my initial framework for handling extensions was largely inadequate, due to its lack of handling of parameters. Negotiation must be possible and the server must be able to return a different (but compatible) set of parameters. (I didn't know that before diving into RFC 7692.)

My version deals with both the client and the server side. It doesn't pass the Autobahn test suite, for reasons I haven't understood yet. Many tests pass, some don't.

RouquinBlanc:
Well if you're on it, that's a great thing!

Instead of trying to push forward my draft, I will have a look at your last code and see if I can spot reasons why autobahn does not pass 👍

aaugustin:
The good news is that we're much more likely to bring this feature to completion if we're two people working on it :-) I'm happy to see that you've make a checklist against the RFC, that's my approach for developing this library.

FYI I like this library but don't have any use case for it anymore, so I'm never sure if the code I write works in real life, beyond test cases and the Autobahn test suite.

Here's a plan -- please let me know if you have other ideas:

- We should compare the two branches and take the best features from both -- likely by copying them from your prototype to my branch, which attempts to cover the entire feature.

- I'd like to rework the extensions negotiation framework in my branch. It doesn't need to be very flexible yet. I'd rather make a release that supports compression and improve negotiation later, as most people won't customize it anyway but would benefit from compression.

- We need to make the Autobahn test suite pass. Help is this area is highly appreciated.  You must modify the configuration (compliance/*.json) like this:

```
   "cases": ["12.*", "13.*"],
   "exclude-cases": [],
```

to test compression. Adjust `cases` to run a subset of tests because they can be quite slow. Once we believe everything works we can do a full run.

`compliance/README` gives some pointers, if you're having trouble let me know and I'll add details.

aaugustin:
Leaving this open for now, but let's follow up in #178!

RouquinBlanc:
I actively use this library, for home project but especially at work! Using it for one year now at least, and still working great ;-) It only lacks compression, which would make a huge difference when talking with mobile devices with low bandwidth...

I just pushed the latest code I was working on (I got carried on after starting!), so that you have my latest things. I just started looking at autobahn test suite but didn't get any pass yet, which is not a surprise as I was more focussing on the "permessage" rather than "deflate". The DEFLATE implementation can be easily inspired by other libraries as well.

Perfectly OK with your plan.

PR #188, opened 2017-05-21
Compressor output should always end with 0x0000FFFF

RouquinBlanc:
According to RFC7692, section 7.2.1
```
 3.  Remove 4 octets (that are 0x00 0x00 0xff 0xff) from the tail end.
       After this step, the last octet of the compressed data contains
       (possibly part of) the DEFLATE header bits with the "BTYPE" bits
       set to 00.
```

### DIFF
--- a/websockets/extensions.py
+++ b/websockets/extensions.py
@@ -96,7 +96,5 @@ class PerMessageDeflate:
             self.encoder.compress(frame.data) +
             self.encoder.flush(zlib.Z_SYNC_FLUSH)
         )
-        if data.endswith(_EMPTY_UNCOMPRESSED_BLOCK):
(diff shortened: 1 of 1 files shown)


aaugustin:
The spec says "append 0x0000FFFF if the message doesn't end with 0x0000FFFF" (step 2) then "strip unconditionally 0x0000FFFF from the end" (step 3, which you're quoting).

This is logically equivalent to "strip 0x0000FFFF from the end", which my implementation does.

I believe the spec is written that way to make it more obvious that appending 0x0000FFFF before decoding doesn't change the contents of the message.

Furthermore, in my testing, the assertion you added isn't always true.

PR #189, opened 2017-05-21
Support `permessage-deflate` parameters

RouquinBlanc:
Implemented a first version of support for extension parameters.

You can force server side minimum parameters when instantiating the PerMessagDeflate class, which will allow users to configure their compression for server side.

Tested on autobahn tests 12.* 13.* while forcing server_max_window_bits=12, client_max_window_bits=10. All green!

### DIFF
--- a/websockets/client.py
+++ b/websockets/client.py
@@ -7,9 +7,8 @@ import asyncio
 import collections.abc
 import email.message
 
-from websockets.extensions import parse_extensions, PerMessageDeflate
-
 from .exceptions import InvalidHandshake, InvalidMessage
+from .extensions import PerMessageDeflate, parse_extensions
 from .handshake import build_request, check_response
 from .http import USER_AGENT, read_response
 from .protocol import CONNECTING, OPEN, WebSocketCommonProtocol
@@ -37,7 +36,9 @@ class WebSocketClientProtocol(WebSocketCommonProtocol):
         self.available_extensions = []
         if use_compression:
             self.available_extensions.append(
-                'permessage-deflate; client_no_context_takeover; client_max_window_bits'
+                'permessage-deflate; '
+                'client_no_context_takeover; '
+                'client_max_window_bits'
             )
         if extensions:
             self.available_extensions.append(extensions)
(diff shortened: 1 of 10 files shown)


RouquinBlanc:
It still needs a closer look at the RFC for conditions when options should be refused, and as well support for fallback configurations; the lib should be able to support things like that:
```
       Sec-WebSocket-Extensions:
           permessage-deflate;
           client_max_window_bits; server_max_window_bits=10,
           permessage-deflate;
           client_max_window_bits
```
Need to put in place some unit tests for extension negotiation!

aaugustin:
This is looking good.

I think the parsing of options in the HTTP header should be extracted to a utility function in the `http` module. It would be very similar to http://werkzeug.pocoo.org/docs/0.12/http/#werkzeug.http.parse_options_header.

I would also create a separate constructor (typically a `from_header` classmethod or similar) for instantiating the extension from the HTTP header. I think it will give the class a better API. In particular it will keep it easier to test.

RouquinBlanc:
As you suggested, I extracted the parsing of the extension header into a separate function, with some initial tests.

I also reworked a bit `PerMessageDeflate.__init__()` to be more generic and work for both server and client. Adding a few extra changes into client.py, the fuzzingserver autobahn tests 12.* and 13.* are now all pass as well!

As for the `from_header` API you suggested, I'm not sure exactly where to put it at the moment; probably into extensions.py, but not into PerMessageDeflate as it would not be generic and difficult to extend later. Probably something aside in that module.

websockets/client.py
```diff
@@ -30,9 +31,17 @@ class WebSocketClientProtocol(WebSocketCommonProtocol):
...
                  origin=None, extensions=None, subprotocols=None,
-                 extra_headers=None, **kwds):
+                 extra_headers=None, use_compression=True, **kwds):
         self.origin = origin
-        self.available_extensions = extensions
+        self.available_extensions = []
+        if use_compression:
+            self.available_extensions.append(
+                'permessage-deflate; '
+                'client_no_context_takeover; '
```
aaugustin: I don't think this is correct, because our implementation retains the sliding window between messages.

aaugustin:
Rebased on top of #178. I'll iterate from there. Thanks a lot.

websockets/client.py
RouquinBlanc: The code I added support both modes, with and without retaining the sliding window. I put that as a default client side config as it is what I get from chrome client or wscat, nothing more. In any case, this will disappear once the code is generalized.

Issue #142, opened 2016-11-15
Worker task not always completed if cancellation happens during close()


aldur:
I am investigating how to close pending tasks on exceptions.
Specifically, I extracted the following snippet of code:

```python
#!/usr/bin/env python3
# -*- coding: utf-8 -*-


import asyncio
import websockets


async def hello():
    async with websockets.connect('wss://echo.websocket.org') as websocket:
        await websocket.send("A" * 1024)
        await websocket.recv()

        raise Exception("Catch me!")

async def runner():
    tasks = [hello() for _ in range(10)]
    done, pending = await asyncio.wait(tasks, return_when=asyncio.FIRST_EXCEPTION)

    if pending:
        for t in pending:
            t.cancel()
        await asyncio.wait(pending)

    for t in done:
        t.exception()  # Retrieve any exception.

asyncio.get_event_loop().run_until_complete(runner())
asyncio.get_event_loop().close()
```

Sometimes, this code results in a series of warnings as follows:

```
task: <Task pending coro=<WebSocketCommonProtocol.run() running at [...]/lib/python3.5/site-packages/websockets/protocol.py:430> wait_for=<Future pending cb=[Task._wakeup()]>>
```

Explicitly calling `close` before raising the exception helps, but does not solve the problem.

How can I manage to cleanly close the pending tasks before handling (i.e. raising) the exception?

aaugustin:
When you say "calling `close()`", do you mean `yield from websocket.close()`?

```python
try:
    # do stuff
finally:
    yield from websocket.close()
```

is supposed to avoid such errors; if they persist it's a bug.

aldur:
It turns out that in the original code I was using `await websocket.close()`, but not in the extracted snippet.

This seems to solve the pending tasks problem, however shouldn't the async context manager guarantee that the websocket is always closed, even on exceptions?

From the docs of `websockets.client.connect`:

```
On Python 3.5, connect() can be used as a asynchronous context manager. In that case, the connection is closed when exiting the context.
```

So why do we need to manually wait for it?

aaugustin:
Right, I missed the async context managed in your example. Indeed it does exactly the same thing as the `try / finally` I proposed in my previous comment.

aaugustin:
Or at least it's supposed to do exactly the same thing.

aldur:
OK so websockets by itself works as supposed, could it be a bug of the context manager?

aaugustin:
[...]

PR #194, opened 2017-07-03
Fix issue #142

cjerdonek:
[Fix for issue #142] This issue was quite tricky for me to develop a test case for. The hardest part is that you need to find a way to call cancel on `close()` when it is at [this particular line of code](https://github.com/aaugustin/websockets/blob/5d361a361a31ce19896dd0fb2c9dc684514e2a09/websockets/protocol.py#L258). I was led to do that in a roundabout way.

I'm open to suggestions on ways to simplify the test or on ways to eliminate the need for sleeps (which could potentially lead to test flakiness, though the sleeps are quite large currently). Or if it's good enough for the time being, that's fine too!

### DIFF
--- a/docs/changelog.rst
+++ b/docs/changelog.rst
@@ -22,6 +22,11 @@ Changelog
   now raises :class:`~websockets.exceptions.InvalidStatus` with a ``code``
   attribute.
 
+* Fixed an issue where
+  :meth:`~websockets.protocol.WebSocketCommonProtocol.close` can return
+  without waiting for an underlying worker task to complete if
+  :meth:`~asyncio.Task.cancel` is called on it.
+
 3.3
 ...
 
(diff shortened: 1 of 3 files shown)


cjerdonek:
FYI, I updated the test case in the PR to have only one call to `asyncio.sleep()` that isn't "minimal." (In contrast, the `asyncio.sleep(1)` call is minimal in the sense that it doesn't actually wait for 1 second but instead waits the minimum amount of time.) The non-minimal call is needed to give `close()` enough time to advance far enough into its execution so that it is actually awaiting on the worker.

cjerdonek:
Ping, @aaugustin. Just wanted to make sure you had seen this. No rush, just hadn't seen an acknowledgment of the diagnosis, etc.

cjerdonek:
Rebased, fixed flake8, and added changelog entry.

cjerdonek:
Rebased.

aaugustin:
Since the bug and the fix are in protocol.py, the test should be in test_protocol.py. Writing a test at a lower level is "easier" (not "easy") because you can control exactly which inputs are fed into the event loop and what happens it what order. See 0ae899ced8e38f1222aaa8dcd2da19a193c7f5c6.

Issue #192, opened 2017-06-30
expose an asynchronous iterator API for consuming messages


cjerdonek:
Given that one of the focuses of `websockets` is simplicity, it seems like it would make sense for `websockets` to expose an API that lets one consume messages using [asynchronous iterators and `async for`](https://www.python.org/dev/peps/pep-0492/#asynchronous-iterators-and-async-for), which were introduced in Python 3.5 via PEP 492.

This would let one replace the following boilerplate consumer pattern everywhere:

    async def handler(websocket, path):
        while True:
            message = await websocket.recv()
            # Do stuff.

with the following simpler and more elegant pattern:

    async def handler(websocket, path):
        async for message in receiving(websocket):
            # Do stuff.

The idea would be that the `websockets` library could expose the asynchronous iterator `receiving()` in the code above (or something very similar).

Python 3.6 lets one define such an iterator very simply, using asynchronous generators which were introduced by [PEP 525](https://www.python.org/dev/peps/pep-0525/).

It would look something like this:

    async def receiving(websocket):
        while True:
            yield await websocket.recv()

(It can also be defined more manually in earlier versions of Python.)

Maybe example code taking advantage of this pattern can be added to the docs even before such an API is added to websockets itself (or even if such an API is never added).

aaugustin:
Yes this looks really cool.

The current API predates the introduction of iterators.

Could we take it as far as:

```
async for message in websocket:
    # do stuff with message
```

That would be damn cool.

cjerdonek:
Thanks for being receptive. Seems likely that could work!

aaugustin:
If you feel like submitting a pull request, perhaps you'll get commit access to merge it yourself :-) (after I review it)

cjerdonek:
What version(s) of Python would the API initially need to support? Would >= Python 3.6 be okay, or is Python 3.5 also necessary?

aaugustin:
Whatever floats your boat :-)

All you need is to document it properly, and make sure the library still imports properly on Python 3.3. Check out the hack for shipping code that uses Python 3.5+ syntax (await / async).

cjerdonek:
I started work on this here using your same pattern / hack: PR #195

aaugustin:
Looks good!

PR #195, opened 2017-07-05
[WIP] Address issue #192: asynchronous iteration

cjerdonek:
This is work in progress (e.g. no tests or documentation yet).

### DIFF
--- a/websockets/py36/protocol.py
+++ b/websockets/py36/protocol.py
@@ -1,0 +1,3 @@
+async def __aiter__(self):
+    while True:
+        yield await self.recv()
--- a/test-iteration.py
+++ b/test-iteration.py
@@ -1,0 +1,23 @@
+import asyncio
+import logging
+import websockets
+
+
+async def demo():
+    async with websockets.connect('ws://echo.websocket.org') as websocket:
+        for i in range(3):
+            await websocket.send('hello!')
+        async for msg in websocket:
+            print('received:', msg)
+
+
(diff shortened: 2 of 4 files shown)


websockets/py36/protocol.py
```diff
@@ -0,0 +1,3 @@
+async def __aiter__(self):
+    while True:
+        yield await self.recv()
```
aaugustin: Perhaps we should catch an exception and exit cleanly if the other end terminates the connection with a 1000 status code?

aaugustin: Or to put it differently: we need to think about the pattern for handling connection termination, in order to avoid boilerplate, such as the need to wrap every `for message in websockets` in a `try/catch` clause.

cjerdonek: Good points, agreed. This was just a POC to get the discussion going, so I hadn't yet given thought myself to proper error handling / cleanup, etc.

cjerdonek:
FYI, so I don't hold up progress on this issue, I just wanted to let you know I'm going to work on some other issues first like #198. The main reason is that I don't have as much intuition for cleanup / error-handling in this case, so I'd want to think about it more carefully before proposing something.

aaugustin:
I'm going to take a look at this. I'm working through PRs, oldest first, now that I finished extensions.

aaugustin:
Tentatively scheduling for 4.0.

Issue #197, opened 2017-07-08
Improve documentation around customizing handling of handshake


cjerdonek:
I was trying to figure out how to return a 404 on a bad path. Two things occur to me that would have made this a bit easier to figure out.

First, the [end of the docstring](https://github.com/aaugustin/websockets/blob/5d361a361a31ce19896dd0fb2c9dc684514e2a09/websockets/server.py#L249) of the server's `handshake()` method could be expanded to say, "Return the URI of the request, _or None if the handshake was aborted_."

Second, the [docstring of the server's `get_response_status()`](https://github.com/aaugustin/websockets/blob/5d361a361a31ce19896dd0fb2c9dc684514e2a09/websockets/server.py#L226) method could be expanded to say that the request path (e.g. to decide whether to return a 404) is available at `self.path` when that method is called.  Currently, you have to notice that `read_http_request()` is called at the beginning of `handshake()` and then see that `read_http_request()` sets `self.path`. But even then, it's not clear if you'd be relying on implementation details.

aaugustin:
Go for it and make a PR!

cjerdonek:
Okay, great. Though if you're open to my comment [here](https://github.com/aaugustin/websockets/issues/116#issuecomment-313893781) (which I wrote later), this issue would probably be subsumed.

aaugustin:
Since #248 was merged, it's time to reopen the discussion here!

aaugustin:
The section about "Port sharing" in deployment.rst should be updated, since it is now possible to implement basic HTTP responses.

aaugustin:
Since I restructured the docs, the best place to put this information would be in its own page, linked from the "howto" section in index.rst.

aaugustin:
By now, I believe this issue has become a duplicate of #354 for which I submitted PR #387.

The original issue is "I was trying to figure out how to return a 404 on a bad path."

The modern API for doing that is `process_request`, like this:

```python
    async def process_request(self, path, request_headers):
        if path != '/':
            return http.HTTPStatus.NOT_FOUND [], b'sorry not found\n'
```

This is close enough to the example in #387.

Issue #116, opened 2016-05-25
WebSockets and HTTP sharing the same port using request headers


pshaughn:
RFC 6455 doesn't mandate that a WebSocket server implementation be able to share the same port with HTTP, but the protocol was designed to allow the possibility.

Currently, WebSocketServerProtocol.handshake doesn't test for the "Upgrade: websocket" header. This doesn't cause any problems because only a WebSocket client would be able to finish the handshake process, but it does appear to rule out port sharing.

It would be convenient in some cases to be able to pass something like an HTTPServer into the WebSocketServer constructor and have requests without the "Upgrade: websocket" header be forwarded to it.

I realize that just putting WebSocket traffic on its own port solves the same problem more easily in most cases, so this idea probably isn't worth implementing if it would take a lot of effort.

aaugustin:
Indeed, I didn't want to implement a HTTP server, only a websocket server (unlike e.g. aiohttp).

You can /probably/ inherit or wrap WebSocketServerProtocol into something that handles HTTP. I made an effort in the design of the API not to rule out this possibility immediately, but I never tried it, so it's entirely possible that it doesn't work...

If you try it, I'm interested into hearing about the results :-) However I suspect it will always be easier to run on a different IP or port.

ghostplant:
I also need this feature, because it will make deployment more easy when been packaged by docker and other engines. Sharing the same port will make it easy to design the port forwarding : )

ghostplant:
I just implement web ssh (https://github.com/ghostplant/websocket-myssh-tunnel.git) using this websockets library, but listening on two ports (8022/8023) seems to be ugly. I am finding a solution.

aaugustin:
I believe the Docker best practice is to run different services in different containers, so you shouldn't serve static contents and websockets from the same container?

Supporting static file isn't enough, someone will ask for dynamic stuff shortly thereafter...

pshaughn:
In your case, ghostplant, it sounds like you can use a regular web-server with server-side proxying, and have it proxy requests on a specific path into the websocket port. Since you're already using Docker, you can easily set up nginx to do that.

ghostplant:
Firstly, Python Flask / Java Tomcat based websocket are designed to share http port and websocket port.

Secondly, if I seperate them two into different docker, when it comes to deploy multiple group of these services, it will be a little more complex, because javascript (from html client side) has to write a complex proceduce to calculate the corrent port of proxy requests to the websocket server.

[...]

PR #199, opened 2017-07-09
Address issue #197 re: handshake customization docs.

cjerdonek:
Addresses #197.

### DIFF
--- a/websockets/server.py
+++ b/websockets/server.py
@@ -218,6 +218,8 @@ class WebSocketServerProtocol(WebSocketCommonProtocol):
         priority = lambda p: client_protos.index(p) + server_protos.index(p)
         return sorted(common_protos, key=priority)[0]
 
+    # This method is declared as a coroutine because overridden versions
+    # can require network requests (e.g. for authentication checks).
     @asyncio.coroutine
     def get_response_status(self, set_header):
         """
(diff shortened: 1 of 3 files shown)


websockets/server.py
```diff
@@ -232,7 +232,9 @@ def get_response_status(self, set_header):
 
         This method may be overridden to check the request headers and set a
         different status, for example to authenticate the request and return
-        ``HTTPStatus.UNAUTHORIZED`` or ``HTTPStatus.FORBIDDEN``.
+        ``HTTPStatus.UNAUTHORIZED`` or ``HTTPStatus.FORBIDDEN``. The
+        current request path can be accessed from within this method
+        as ``self.path``.
```
aaugustin: It may be useful to describe how to access headers (``request_headers``, ``raw_request_headers``) as well, which doesn't quite belong to a docstring.

Perhaps it would be best to add a section to the docs about customizing the handshake process and give all the details there. What do you think?

(I also need to look at the other related issue.)

cjerdonek: A section like that in the docs makes sense, though I think I'd like to wait until after #116 is resolved to do that since it may require rewriting a portion those docs.

I do think the fact that `self.path` and `self.request_headers` are guaranteed to be available should be stated in the same place developers are told the method is overridable (currently stated in the docstring). That way implementers are clearer on what they can / should be relying on when they override.

cjerdonek: Note to self: `self.origin` is also available.

cjerdonek:
Patch updated based on feedback.

aaugustin:
This looks pretty good, thanks a lot for your work.

I'll take the time to review next week-end (hopefully).

cjerdonek:
Rebased after PR #236.

aaugustin:
This needs updating after #248.

aaugustin:
I'm closing this PR because it's out of date. The corresponding issue #197 is still open.

PR #200, opened 2017-07-09
Replace MIME parsing with custom HTTP parsing.

### LINKED ISSUE #19: Consider using a proper HTTP implementation
The HTTP parsing and generation code is very simple at this time.

It could be worth relying on an external library for these parts.

aaugustin:
Given that websockets makes straightforward use of HTTP, that websocket
implementations can be expected not to exhibit legacy behaviors, and
that RFC 7230 deprecates this behavior, parsing HTTP is doable.

Thanks https://github.com/njsmith/h11 for providing some inspiration,
especially for translating the RFC to regular expressions and figuring
out some edge cases.

I expect the new implementation to be faster, since it has a much
tighter focus than the stdlib's general purpose MIME parser, and
possibly more secure, since it was written from the beginning with
security as a primary goal (with the caveat that it's new code,
which means it's more likely to have security issues).

Fix #19.

### DIFF
--- a/websockets/client.py
+++ b/websockets/client.py
@@ -5,11 +5,10 @@ The :mod:`websockets.client` module defines a simple WebSocket client API.
 
 import asyncio
 import collections.abc
-import http.client
 
 from .exceptions import InvalidHandshake, InvalidMessage
 from .handshake import build_request, check_response
-from .http import USER_AGENT, read_response
+from .http import USER_AGENT, build_headers, read_response
 from .protocol import CONNECTING, OPEN, WebSocketCommonProtocol
 from .uri import parse_uri
 
@@ -35,8 +34,7 @@ class WebSocketClientProtocol(WebSocketCommonProtocol):
 
         """
         self.path = path
-        self.request_headers = http.client.HTTPMessage()
-        self.request_headers._headers = headers     # HACK
+        self.request_headers = build_headers(headers)
         self.raw_request_headers = headers
 
         # Since the path and headers only contain ASCII characters,
@@ -62,8 +60,7 @@ class WebSocketClientProtocol(WebSocketCommonProtocol):
         except ValueError as exc:
             raise InvalidMessage("Malformed HTTP message") from exc
 
-        self.response_headers = http.client.HTTPMessage()
-        self.response_headers._headers = headers    # HACK
+        self.response_headers = build_headers(headers)
         self.raw_response_headers = headers
 
         return status_code, self.response_headers
--- a/websockets/client.py
+++ b/websockets/client.py
@@ -5,7 +5,7 @@ The :mod:`websockets.client` module defines a simple WebSocket client API.
 
 import asyncio
 import collections.abc
-import email.message
+import http.client
 
 from .exceptions import InvalidHandshake, InvalidMessage
 from .handshake import build_request, check_response
@@ -35,9 +35,8 @@ class WebSocketClientProtocol(WebSocketCommonProtocol):
 
         """
         self.path = path
-        self.request_headers = email.message.Message()
-        for name, value in headers:
-            self.request_headers[name] = value
+        self.request_headers = http.client.HTTPMessage()
+        self.request_headers._headers = headers     # HACK
         self.raw_request_headers = headers
 
         # Since the path and headers only contain ASCII characters,
@@ -63,10 +62,11 @@ class WebSocketClientProtocol(WebSocketCommonProtocol):
         except ValueError as exc:
             raise InvalidMessage("Malformed HTTP message") from exc
 
-        self.response_headers = headers
-        self.raw_response_headers = list(headers.raw_items())
+        self.response_headers = http.client.HTTPMessage()
+        self.response_headers._headers = headers    # HACK
+        self.raw_response_headers = headers
 
-        return status_code, headers
+        return status_code, self.response_headers
 
     def process_subprotocol(self, get_header, subprotocols=None):
         """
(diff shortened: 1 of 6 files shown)


aaugustin:
@cjerdonek Any thoughts on this?

aaugustin:
Context: I got side-tracked and cleaned this up because I was wondering how to handle multiple `Sec-Websocket-Extension` headers. 

🐰 🕳

websockets/client.py
```diff
@@ -35,9 +35,8 @@ def write_http_request(self, path, headers):
 
         """
         self.path = path
-        self.request_headers = email.message.Message()
-        for name, value in headers:
-            self.request_headers[name] = value
+        self.request_headers = http.client.HTTPMessage()
+        self.request_headers._headers = headers     # HACK
```
cjerdonek: I looked at CPython's code [here](https://github.com/python/cpython/blob/9648088e6ccd6d0cc04f450f55628fd8eda3784c/Lib/http/client.py#L163), and HTTPMessage doesn't really add much, and it's only reluctantly part of the public API, if that.

What about the idea of defining a `websockets.HTTPMessage(email.message.Message)` class?  At the least, this would let you DRY up the four occurrences of:

    self.response_headers._headers = headers    # HACK

and confine the hack to one spot.  You could also add a custom property / method on the class to eliminate the need for separate `self.raw_*_headers` attributes on the `WebSocketCommonProtocol` class.

Also, it might be worth having a couple unit tests to check that setting `_headers` directly will continue to work in future Pythons -- giving you the same result as going through `email.message.Message`'s (slightly slower) public API, etc.

cjerdonek: These two frequently used functions could also go on the class.

    set_header = lambda k, v: headers.append((k, v))
    get_header = lambda k: headers.get(k, '')

aaugustin: Yes, I ended up using `http.client.HTTPMessage` a bit reluctantly and I'm not very happy with that hack.

The constraints here are:

- not losing any information: I believe the the `raw_*_headers` lists of `(name, value)` pairs are a good way to achieve that; I'm happy with them (even though it's unlikely that they will be used widely)
- providing a convenient API for accessing headers: we're immediately getting into the problem of multi-value dicts where 99,99% of accesses expect a single value; each library tends to invent its own API; I decided to use the closest matching data structure in the standard library to avoid adding one more variant to the Python ecosystem
- minimizing performance overhead; the HACK is the only way I found to instantiate a `HTTPMessage` without running a lot of code; it don't expect the implementation to change, but the risk exists in theory; I believe tests cover this
- not reinventing the wheel: not only does `http.client.HTTPMessage` feel semantically correct for representing HTTP headers, but it's backwards compatible because it inherits `email.message.Message` (the less semantically correct type previously used there); see also https://docs.python.org/3/library/http.client.html#http.client.HTTPResponse.msg

I considered using properties and rejected that idea because it would make the code more verbose, slightly less efficient, and make it less obvious where these values are set.

aaugustin: Conversely, defining get/set_header at the class level is certainly more efficient that redefining them dynamically with each execution.

Historically they couldn't be methods because headers weren't stored as attributes. Certainly a good time to reconsider that part of the design.

websockets/http.py
```diff
@@ -34,20 +53,38 @@ def read_request(stream):
     ``stream`` is an :class:`~asyncio.StreamReader`.
 
     Return ``(path, headers)`` where ``path`` is a :class:`str` and
-    ``headers`` is a :class:`~email.message.Message`. ``path`` isn't
-    URL-decoded.
+    ``headers`` is a list of ``(name, value)`` tuples.
```
aaugustin: I could, but I like using basic Python types in low level APIs when possible. A list of pairs is pretty manageable.

aaugustin:
Thanks for the review, I'll make changes next week-end.

websockets/client.py
cjerdonek: To start with, I think it could be as simple as a class with two attributes for "raw" and "not raw" (matching the two attributes currently stored on the Protocol classes), and you wouldn't lose anything. And then you could add whatever helper methods as needed, like the get / set header functions.

Later on (and only if you wanted), you could get more sophisticated with how you store and mutate the raw header info internally to the class.

cjerdonek:
Serious question: since the library would now be doing its own parsing, does it even need to construct and set an `HTTPMessage` object?

It looks like the only place the class is actually being used (aside from test code) is to call its `get()` method in the following two spots.

For the server, [here](https://github.com/aaugustin/websockets/blob/b1d09a1b1c36215a59f6a630215ed15d4582b3a8/websockets/server.py#L270) in its `handshake()` method:

    get_header = lambda k: headers.get(k, '')

And in the client, [here](https://github.com/aaugustin/websockets/blob/b1d09a1b1c36215a59f6a630215ed15d4582b3a8/websockets/client.py#L121) also in its `handshake()` method:

    get_header = lambda k: headers.get(k, '')

aaugustin:
Headers represented as a MIME message are currently part of the public API, but I'm open to changing that, see #210.

aaugustin:
I added a commit that wraps the `# HACK` into a helper function.

I looked at refactoring `get/set_header` as methods, but I like that the `path`, `request_headers` and `raw_request_headers` attributes are set on instances atomically when sending out the HTTP request, so I didn't make that change for now.

cjerdonek:
Thanks for making a helper function and for opening issue #210. I appreciate it.

PR #154, opened 2017-01-15
Refactor handshake for extensibility.

aaugustin:

PR #202, opened 2017-07-09
Address issue #116: make existing hook take place before the handshake

cjerdonek:
This is an illustration of what I proposed in [this comment](https://github.com/aaugustin/websockets/issues/116#issuecomment-313893781) to add a "pre-handshake" hook. All of the tests still pass (at least on my machine with 3.6.1).

If this approach is acceptable, I could make needed documentation changes, etc, and add a test of using the hook.

### DIFF
--- a/websockets/server.py
+++ b/websockets/server.py
@@ -66,7 +66,8 @@ class WebSocketServerProtocol(WebSocketCommonProtocol):
         # it attemps to log relevant ones and close the connection properly.
         try:
             try:
-                response_headers = yield from self.pre_handshake(origins=self.origins)
+                response_headers = (
+                    yield from self.pre_handshake(origins=self.origins))
                 if response_headers is None:
                     # Then the response was already written.
                     return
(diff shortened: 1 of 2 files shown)


cjerdonek:
Okay, I added a test and also checked that the test in PR #208 passes, too. For the documentation changes (which would be small), it would be best to wait until after PR #199 is committed IMO. Otherwise, it will create duplicated work.

cjerdonek:
Rebased (still need to address coverage).

cjerdonek:
Rebased, and brought coverage back to 100%.

cjerdonek:
Rebased.

cjerdonek:
Rebased after PR #236.

aaugustin:
Thanks for highlighting the drawbacks of #154. Indeed the hook must be before `check_request()`.

I'm proposing a slightly more straightforward approach in #248, which merges `get_status_response()` and `pre_handshake()` into a single extension point.

Tentatively closing in favor of #248.

Issue #207, opened 2017-07-14
server's handshake() method does not always set self.origin


cjerdonek:
In the process of writing a test for PR #202, I noticed that the server's `handshake()` method doesn't set `self.origin` if `origins` isn't passed. The line of code is [here](https://github.com/aaugustin/websockets/blob/b1d09a1b1c36215a59f6a630215ed15d4582b3a8/websockets/server.py#L274), and it's because `process_origin()` [doesn't return a value](https://github.com/aaugustin/websockets/blob/b1d09a1b1c36215a59f6a630215ed15d4582b3a8/websockets/server.py#L193) if `origins` is None.

The main negative consequence of this is that `self.origin` isn't always available from within `get_response_status()`.

cjerdonek:
I posted PR #208 for this.

cjerdonek:
Addressed via PR #208.

PR #208, opened 2017-07-14
Fix issue #207: always set self.origin

cjerdonek:
This fixes issue #207.

It would be good if this could be committed before related PR's so the test can be present for the other PR's.

### DIFF
--- a/websockets/test_client_server.py
+++ b/websockets/test_client_server.py
@@ -1,4 +1,5 @@
 import asyncio
+import email.message
 import http
 import logging
 import os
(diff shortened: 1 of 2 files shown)


cjerdonek:
Note that the test added here is useful independent of issue #207. For example, it checks some of what is documented in PR #199, namely the availability of the attributes it lists as being available from within `get_response_status()`.

websockets/test_client_server.py
```diff
@@ -217,6 +218,35 @@ def test_protocol_custom_response_headers_list(self):
...
+        class SaveAttributesProtocol(WebSocketServerProtocol):
+            @asyncio.coroutine
+            def get_response_status(self, set_header):
+                attrs['origin'] = self.origin
+                attrs['path'] = self.path
+                attrs['raw_request_headers'] = self.raw_request_headers.copy()
+                attrs['request_headers'] = self.request_headers
+                status = yield from super().get_response_status(set_header)
+                return status
+        self.start_server(klass=SaveAttributesProtocol)
```
aaugustin: Some extra newlines would help readability.

websockets/test_client_server.py
```diff
@@ -217,6 +218,35 @@ def test_protocol_custom_response_headers_list(self):
...
+        try:
+            self.start_client(path='foo/bar', origin='http://otherhost')
+            self.assertEqual(attrs['origin'], 'http://otherhost')
+            self.assertEqual(attrs['path'], '/foo/bar')
+            # To reduce test brittleness, only check one nontrivial aspect
+            # of the request headers.
+            self.assertTrue(('Origin', 'http://otherhost') in
```
aaugustin: `self.assertIn(('Origin', 'http://otherhost'), attrs['raw_request_headers'])`

websockets/test_client_server.py
```diff
@@ -217,6 +218,35 @@ def test_protocol_custom_response_headers_list(self):
...
+                attrs['raw_request_headers'])
+            request_headers = attrs['request_headers']
+            self.assertTrue(isinstance(request_headers, email.message.Message))
```
aaugustin: `self.assertIsInstance(request_headers, http.client.HTTPMessage)`

(I changed the type a few minutes ago.)

cjerdonek:
Thanks for the review! I can make those changes and commit.

Question: when it comes to committing to master, do you have a preference between (1) "squashing" and committing (with no merge commit), and (2) not squashing and merging the PR commits?

cjerdonek:
Addressed review comments, squashed, and committed as e0915479ce582b0b69b551cc62ab2eb4e6cd39f9.

aaugustin:
I prefer squashing when the history isn't clean (e.g. lots of small fixes until tests pass). If the PR has several logical commits, performing a rebase or a merge is fine. I don't care very much ;-)

Issue #198, opened 2017-07-09
provide programmatic access to status code if client connection fails


cjerdonek:
Currently, if [`client.connect()`](https://github.com/aaugustin/websockets/blob/5d361a361a31ce19896dd0fb2c9dc684514e2a09/websockets/client.py#L136) fails, there doesn't seem to be a programmatic way to get the HTTP status code when applicable.  For example, in the case of a 404, the following traceback occurs. The status code occurs in the error text but not otherwise:

    Traceback (most recent call last):
      ...
      File "/.../lib/python3.6/site-packages/websockets/py35/client.py", line 12, in __aenter__
        self.websocket = await self
      File "/.../lib/python3.6/site-packages/websockets/py35/client.py", line 19, in __await__
        return (yield from self.client)
      File "/.../lib/python3.6/site-packages/websockets/client.py", line 165, in connect
        extra_headers=extra_headers)
      File "/.../lib/python3.6/site-packages/websockets/client.py", line 83, in handshake
        raise InvalidHandshake("Bad status code: {}".format(status_code))
    websockets.exceptions.InvalidHandshake: Bad status code: 404

The relevant line of code is [here](https://github.com/aaugustin/websockets/blob/5d361a361a31ce19896dd0fb2c9dc684514e2a09/websockets/client.py#L124).

One possible solution would be to define an `HttpError` exception class that has a `status_code` attribute. This would be similar to [`urllib.error`'s `HTTPError` class](https://docs.python.org/3.6/library/urllib.error.html#urllib.error.HTTPError).

aaugustin:
Yes, that part of the API isn't convenient. Your proposal works for me.

cjerdonek:
I posted PR #209 for this.

PR #209, opened 2017-07-15
Address issue #198: expose invalid status code during connection.

cjerdonek:
This addresses issue #198.


websockets/exceptions.py
```diff
@@ -26,13 +26,31 @@ class InvalidOrigin(InvalidHandshake):
...
+
+
 class InvalidState(Exception):
     """
     Exception raised when an operation is forbidden in the current state.
 
     """
 
 
+# TODO: DRY up with InvalidStatus?
```
aaugustin: `InvalidStatus` stores a HTTP status code (three digits). `ConnectionClosed` stores a websocket status code (four digits). That makes them different types — even if we represent both as integers because we don't bother too much with types in Python. For this reason, I don't think we should merge these two exceptions types.

aaugustin: You can remove the comment if you agree with that.

websockets/test_client_server.py
```diff
@@ -401,6 +402,16 @@ def test_server_shuts_down_during_connection_handling(self):
         # Websocket connection terminates with 1001 Going Away.
         self.assertEqual(self.client.close_code, 1001)
 
+    def test_invalid_status_error_during_client_connect(self):
+        self.start_server(klass=ForbiddenWebSocketServerProtocol)
+        with self.assertRaises(InvalidStatus) as context:
```
aaugustin: `as raised`, for consistency with the test just above?

websockets/exceptions.py
cjerdonek: Sounds good, and thanks for the explanation.

cjerdonek: Actually, looking back at this, the other thought I had in mind (in addition to the possibility of subclassing) was to DRY up the code responsible for constructing the message attribute (in which case the more appropriate location for the TODO would be in the method body). This was for aesthetic reasons to ensure the message format would be identical for the user, and for normal code reasons. In InvalidStatus, I used the format: `f"{description}: {code} ({reason})"`.

Whether or not you agree with the change (which could be done in a separate PR), I can remove the TODO. It was mainly to spark the conversation we are currently having as to whether code should be shared. :)

websockets/test_client_server.py
cjerdonek: Sure. Note that `context` is also used once in test_protocol.py. Incidentally, the [Python docs](https://docs.python.org/3/library/unittest.html#unittest.TestCase.assertRaises) use `cm` in their example code snippet.

cjerdonek:
Okay, I rebased. A couple comments though (one of which I need your feedback on):

First, I added a new test module (`test_exceptions.py`) to keep (branch) code coverage at 100%. This seemed like the easiest thing to do.

Second, how would you like to handle converting the "reason" from a bytestring to string. I'm temporarily doing this [here](https://github.com/aaugustin/websockets/pull/209/commits/e8a2dd0023736ac7e463e5c9229fb0504fc99d98#diff-2ace2674b5888c98b13c496efa2903fdR131). My thought is that something like this should be okay since UnicodeDecodeError is a subclass of ValueError. There should be a test though.

cjerdonek:
On second thought, the [portion of the spec](https://tools.ietf.org/html/rfc7230#section-3.1.2) you link to in the code says--

    A client SHOULD ignore the reason-phrase content.

      reason-phrase  = *( HTAB / SP / VCHAR / obs-text )

So I'm inclined not to return the reason after all. That also has the benefit of being simplest.

By the way, I see the code does this [just prior](https://github.com/aaugustin/websockets/blob/e0915479ce582b0b69b551cc62ab2eb4e6cd39f9/websockets/http.py#L126):

    if not _value_re.match(reason):
        raise ValueError("Invalid HTTP reason phrase: %r" % reason)

So it seems like it would be more correct to remove that.

cjerdonek:
PR updated.

aaugustin:
`test_exceptions.py` is good. Although this isn't enforced, I'm trying to make sure that `test_foo.py` brings 100% branch coverage on `foo.py`. There's an exception to this rule for `client.py` and `server.py` because they're tested together, against each other.

Indeed, the reason phrase doesn't matter very much. For protocol compliance and security we must fail on invalid characters so I'd like to keep that check.

If you need to convert it non-destructively from `bytes` to `str`, use  `reason.decode('ascii', 'surrogateescape')`. That will do the job and never raise an exception (I think). I expect it to be ASCII in most cases. But that point seems moot now :-)

cjerdonek:
Thanks! And yes, later on I saw elsewhere in your code you used `'surrogateescape'` for the same thing.

cjerdonek:
By the way, do you know if there's an easy way to make the coverage invocation in tox.ini report the line numbers responsible for failing to reach 100%? IIRC, it doesn't do this, and it's not always obvious.

cjerdonek:
I answered my own question here: https://github.com/aaugustin/websockets/issues/212

Issue #215, opened 2017-07-22
ConnectionError handling in WebSocketServerProtocol.handler() not tested


cjerdonek:
When I comment out [this code](https://github.com/aaugustin/websockets/blob/93f258fb455d68ec931a53dde3d138e27ed5ef72/websockets/server.py#L73) in `WebSocketServerProtocol.handler()`, the tests still pass:

    try:
        path = yield from self.handshake(
            origins=self.origins, subprotocols=self.subprotocols,
            extra_headers=self.extra_headers)
    # except ConnectionError as exc:
    #     logger.debug(
    #         "Connection error in opening handshake", exc_info=True)
    #     raise
    except Exception as exc:
        if self._is_server_shutting_down(exc):
            response = ('HTTP/1.1 503 Service Unavailable\r\n\r\n'
                        'Server is shutting down.')
        elif isinstance(exc, InvalidOrigin):
            response = 'HTTP/1.1 403 Forbidden\r\n\r\n' + str(exc)
        elif isinstance(exc, InvalidHandshake):
            response = 'HTTP/1.1 400 Bad Request\r\n\r\n' + str(exc)
        else:
            logger.warning("Error in opening handshake", exc_info=True)
            response = ('HTTP/1.1 500 Internal Server Error\r\n\r\n'
                        'See server log for more information.')
        self.writer.write(response.encode())
        raise

Is handling `ConnectionError` actually needed here? If so, the desired behavior should be tested.

I noticed this when I was working on PR #202.

cjerdonek:
I added PR #217 for this.

aaugustin:
Hu, right, I was sloppy on tests... I could get away with this until you started double-checking my work :-)

cjerdonek:
:)

aaugustin:
I took the time to review the PR this morning and it occurred to me that the test logic could be simplified.

The `WriteCollectingProtocol` in #217 does three things:

1. inject an arbitrary exception in `handshake()` (which was previously achieved with a `@unittest.mock.patch`)
2. track writes made by the protocol
3. track the exception raised by the client when it starts

Another `@unittest.mock.patch` allows handling 2. more consistently with the remainer of the file, as shown in #280.

#217 also added a test for a situation that is already tested `test_server_error_in_handshake`. It's best not to test two different situations in a single test method.

PR #217, opened 2017-07-23
Address issue #215: test server ConnectionError handling.

cjerdonek:
This addresses issue #215.

### DIFF
--- a/websockets/test_client_server.py
+++ b/websockets/test_client_server.py
@@ -440,16 +440,39 @@ class ClientServerTests(unittest.TestCase):
         self.assertEqual(str(exception), "Status code not 101: 403")
         self.assertEqual(exception.code, 403)
 
-    @with_server()
-    @unittest.mock.patch('websockets.server.read_request')
-    def test_connection_error_during_opening_handshake(self, _read_request):
-        _read_request.side_effect = ConnectionError
+    def do_error_during_opening_handshake(self, exception_cls):
+        written = []
 
-        # Exception appears to be platform-dependent: InvalidHandshake on
-        # macOS, ConnectionResetError on Linux. This doesn't matter; this
-        # test primarily aims at covering a code path on the server side.
-        with self.assertRaises(Exception):
-            self.start_client()
+        class WriteCollectingProtocol(WebSocketServerProtocol):
+
+            def client_connected(self, reader, writer):
+                original_write = writer.write
+
+                def collecting_write(data):
+                    original_write(data)
+                    written.append(data)
+                writer.write = collecting_write
+                super().client_connected(reader, writer)
+
+            @asyncio.coroutine
+            def handshake(self, *args, **kwargs):
+                raise exception_cls()
+
+        with self.temp_server(klass=WriteCollectingProtocol):
+            # Exception appears to be platform-dependent: InvalidHandshake on
+            # macOS, ConnectionResetError on Linux. This doesn't matter; this
+            # test primarily aims at covering a code path on the server side.
+            with self.assertRaises(Exception):
+                self.start_client()
+
+        return written
+
+    def test_connection_error_during_opening_handshake(self):
+        written = self.do_error_during_opening_handshake(Exception)
+        self.assertTrue(written)
+
+        written = self.do_error_during_opening_handshake(ConnectionError)
+        self.assertFalse(written)
 
     @with_server()
     @unittest.mock.patch('websockets.server.WebSocketServerProtocol.close')
--- a/websockets/test_client_server.py
+++ b/websockets/test_client_server.py
@@ -473,7 +473,7 @@ class ClientServerTests(unittest.TestCase):
 
         written, exception = self.raise_while_opening(ConnectionError)
         self.assertFalse(written)
-        # This exception is currently platform-dependent.  It was observed
+        # This exception is currently platform-dependent. It was observed
         # to be ConnectionResetError on Linux in the non-SSL case, and
         # InvalidMessage otherwise (including both Linux and Mac OS X).
         # This doesn't matter though since this test is primarily for testing
--- a/websockets/test_client_server.py
+++ b/websockets/test_client_server.py
@@ -10,7 +10,8 @@ import unittest.mock
 from contextlib import contextmanager
 
 from .client import *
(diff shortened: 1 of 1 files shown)


websockets/test_client_server.py
```diff
@@ -442,18 +442,41 @@ def test_invalid_status_error_during_client_connect(self):
...
+                super().client_connected(reader, writer)
+
+            @asyncio.coroutine
+            def handshake(self, *args, **kwargs):
+                raise exception_cls()
+
+        self.start_server(klass=WriteCollectingProtocol)
         # Exception appears to be platform-dependent: InvalidHandshake on
         # macOS, ConnectionResetError on Linux. This doesn't matter; this
         # test primarily aims at covering a code path on the server side.
```
aaugustin: Here in this comment, I'm saying that I only wrote this test to get 100% code coverage.

I think I wrote the comment after noticing that tests were passing locally on my mac but failing on the CI service.

You improved the test to actually test the behavior instead of going through the code. I think we should re-check the exception type and perhaps update the comment.

aaugustin: FWIW I modified the following lines like this:
```python
        with self.assertRaises(Exception) as exc:
            self.start_client()
        print(exc.exception)
```
and I'm getting these exceptions:
```
Status code not 101: 500
Malformed HTTP message
```
on macOS.

cjerdonek: Good suggestion. On inspection, it seems like we fully control those exception types on the client side (and so may be able to tighten the assertion), but I can't easily confirm on Linux as I use a Mac, too (and I don't know what the code did when the comment was written).

Really, what this test is checking is whether [this line](https://github.com/aaugustin/websockets/blob/ccc5d34f068e1e3e6b51c49aa0c69983d9cd104d/websockets/server.py#L89) is getting called on the server side:

    self.writer.write(response.encode())

 which is the only difference between the `ConnectionError` case and the non-`ConnectionError` case, aside from debug vs. warning logging levels. Can you confirm this is really how you want the code to behave?

Beyond that, like you say in the comment, the test is more for the server side, so checking the exception type on the client side seems more of a can't hurt / nice to be there. If we're interested in the client-side code path, it might deserve to be separated out into its own test if the code path isn't already covered. Or maybe there is a unique code path on the client for this case, and this is the easiest way to trigger it.

aaugustin: You can fix the merge conflicts, add some debug code, and look in the Circle CI output to check the behavior on Linux.

This isn't very convenient because every iteration takes two or three minutes, but it's bearable for small checks.

It's late here, I'll investigate the remainder of your comment another day!

cjerdonek: I don't think I've seen Circle CI runs on anything other than master (i.e. not for PR runs). Is there a way to see and/or trigger those? Maybe I'm just not looking in the right place. Or were you suggesting to merge the test to master with the debug code and looking there.

cjerdonek: Oh, maybe it's just a matter of pushing the branch to this repo. I haven't tried that yet!

cjerdonek:
Rebased. Will do the debugging separately.

cjerdonek:
Okay, here are the [results](https://circleci.com/gh/aaugustin/websockets/75) (same for all Pythons):

    ClientServerTests

    Exception -> InvalidStatus('Status code not 101: 500',)
    ConnectionError -> ConnectionResetError(104, 'Connection reset by peer')

    SSLClientServerTests

    Exception -> InvalidStatus('Status code not 101: 500',)
    ConnectionError -> InvalidMessage('Malformed HTTP message',)

The exception on the left is the exception type the test raises inside `server.handshake()`, and the right is the exception on the client. What's interesting to me is that they're different depending on SSL. Is that desired?

aaugustin:
Yes, there's a difference in connection termination with SSL, although I don't know exactly the reason. It isn't related to websockets; it's an abstraction leak due to the SSL encapsulation. There are some special cases in the asyncio source code related to this. That's about all I know :-|

> Can you confirm this is really how you want the code to behave?

1. There's no point trying to write to the network after the connection has broken, since it would only cause a further exception (perhaps after a timeout); that's why the code doesn't answer the handshake request in that case. I think that's correct.
2. Connections break down all the time, sometimes while sending the HTTP headers of the handshake request. Since nothing very interesting has happened at this point, logging such errors merely creates noise. However I'm keeping logs at the "debug" level because people occasionally need them to debug tricky errors. That's #126 (and also a few other tickets, I think).
3. In this test, I'm injecting ConnectionError when the server attempts to read the handshake request, but the network connection isn't actually broken: the client still listens for a response to its handshake request! Clearly I wrote the test that way because I wanted to cover the `except ConnectionError` clause on the server. So the situation is unrealistic for the client anyway; it's acceptable if we don't specify exactly what happens on the client in that case.

Eventually the connection is closed without response. Apparently, with SSL, read() returns "here are 0 bytes", while without SSL, it returns "hey the connection was closed!", signalling EOF in both cases. Maybe that's because the SSL layer catches the "connection close" event and cleans things up (I'm speculating wildly here...)

Having different exception types isn't exactly "desired". It might cause issues if devs work withouth SSL locally and deploy with SSL. But in the grand scheme of things, I can live with it.

cjerdonek:
Great, thank you for the thorough explanation. I can update the patch with more specific exception types. For the `ConnectionError` case, would you prefer that I say `ConnectionResetError` and `InvalidMessage` are acceptable for both SSL and non-SSL, or should I restrict the dual possibilities just to the non-SSL case?

aaugustin:
Don't bother restricting the possibilities.

If we wanted to be more specific, we'd have to figure out how we're closing the connection on the server side in this scenario. Then we could talk about the expected behavior on the client side. I don't think it's worth the effort!

cjerdonek:
Okay, I updated the patch per your comments. I think I interpreted your last comment correctly ("restrict" can be interpreted a couple ways). If not, let me know!

aaugustin:
I'm planning to rebase this PR — which I neglected for way too long, sorry — then review and hopefully merge it.

Once that's done I'll release version 4.0.

aaugustin:
Superseded by #280.

Issue #223, opened 2017-07-23
Invalid read of mask in speedups.c


mayeut:
mask is read as follow on non SSE2 platforms:
`uint64_t mask_64 = (*(uint64_t *)mask << 32) | *(uint64_t *)mask;`

mask contains only 4 values. Even if this works properly on little endian platforms (unlikely to read outside of mapped memory), tests from `test_utils.py` are failing on big endian platforms (tested on fedora 24 ppc64).

PR #224, opened 2017-07-23
Correctly read mask in non SSE2 cases in speedups.c

mayeut:
Fixes aaugustin/websockets#223


aaugustin:
Ha that was dumb. Thank you.

PR #226, opened 2017-07-23
Replace modulo on signed integers by masking

mayeut:
Compiler is not smart enough to know that both operands are positive &
optimize this with masking. We do what it should have been doing.
This is not critical given that only the remainder (at most 7 bytes) is impacted but nevertheless.

### DIFF
--- a/websockets/speedups.c
+++ b/websockets/speedups.c
@@ -8,7 +8,7 @@
 #include <emmintrin.h>
 #endif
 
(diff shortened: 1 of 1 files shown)


aaugustin:
I assumed the compiler would figure that out... Py_ssize_t must be signed.

mayeut:
Indeed Py_ssize_t is signed.

PR #227, opened 2017-07-23
Don't accept Unicode objects in speedups.apply_mask

mayeut:
Only non-mutable bytes-like objects shall be supported


aaugustin:
I thought about this yesterday and decided against it, but I can remember why. Let's do it.

Issue #221, opened 2017-07-23
Allow custom arguments to EventLoop.create_connection to support proxies / custom transport


FraMecca:
As specified [here](https://docs.python.org/3/library/asyncio-eventloop.html#creating-connections) loop.create_connection (of client.py:191) supports a custom socket for transport.

This would allow, for example, websockets over TOR provided that I open a socket to a TOR domain indipendently from your library.

Even if you provide **kwds in the function call, it is not valid because as the docs specifies if a socket is given explicitely,  none of host, port, family, proto, flags and local_addr should be specified.

If you want I can provide a possible workaround but I don't know about the other legitimate use cases of **kwds in loop.create_connection.

aaugustin:
Thanks for the bug report.

Would you like to try submitting a pull request?

I believe it should look like this:

```
host = wsuri.host if kwds.get('sock') is None else None
port = wsuri.port if kwds.get('sock') is None else None
transport, protocol = yield from loop.create_connection(
    factory, host, port, **kwds)
```

I'm not really in a position to test this change, but if it works for you, I'm happy to merge it.

PR #229, opened 2017-07-23
Allow custom arguments to EventLoop.create_connection to support proxies / custom transport 

FraMecca:
As of #221

### DIFF
--- a/websockets/test_client_server.py
+++ b/websockets/test_client_server.py
@@ -467,7 +467,7 @@ class ClientServerTests(unittest.TestCase):
         self.assertEqual(self.client.close_code, 1006)
 
 
-class CustomSocketClientTests(ClientServerTests):
+class CustomSocketClientTests(unittest.TestCase):
     class CustomSocket(socket.socket):
         def __init__(self, typ, transp):
             self.used = False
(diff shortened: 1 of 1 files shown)


aaugustin:
Good. Now you need a test to ensure coverage remains at 100%. Most likely it should go in test_client_server.py.

FraMecca:
Where is test_client_server.py located? I can't find it.

aaugustin:
https://github.com/aaugustin/websockets/blob/master/websockets/test_client_server.py

FraMecca:
I have implemented a test case just by overriding the start_client function in a new test class and repeating all the test for that class using a socket opened using the socket library.

My test class also checks that the socket has effectively been used

aaugustin:
Testing that the socket is actually used is a good idea.

That said I don't think it's a good trade off to duplicate all tests for this because the test suite will take significantly longer. Could you refactor this into a single test?

Then make sure the code matches the coding style (`flake8 websockets`) and I think this will be good to merge.

Thanks!

FraMecca:
The total of the tests takes less than two seconds in my machine. 
Also, I may have explained badly what I did. 

My test class inherits ClientServerTests and replicates all of the class testes except for the one closing the connection before handshake.

There are many that are not needed, for example the ones testing the various headers, but I didn't bother because on my machine my test class has a overhead of 0.2 seconds, bringing the test time from an average of 1.7 seconds to 1.9 seconds.

If you prefer, I may go and examine each test of the ClientServerTest class and remove the ones that are not needed.

FraMecca:
Also, I corrected the code style

FraMecca:
Ok, sounds reasonable.
I'll leave for some days, I'll get to it as soon as I am back home

aaugustin:
Thanks and no hurry!

FraMecca:
Ok, I have reduced the test size using only the test I consider useful for my case.

You should consider using custom rules for flake8 linting because they are too strict (see line 501, client has to be renamed to clnt to avoid >80 chars lines).

aaugustin:
I tried to clean this up in #242 but didn't get to something fully functional.

* I squashed the commits
* I improved the implementation a bit: I kept it functional if the function is called with an explicit `sock=None` parameter)
* I simplified the tests: I believe this needs exactly one test
* I made the test run with and without SSL, which is the interesting variant, as discussed earlier, and which shows a problem with the current implementation:

`ValueError: You must set server_hostname when using ssl without a host`

Since I don't know the use case for a custom `sock` parameter, I'm not sure if we should require users to provide `server_hostname` — and modify the test accordingly — or determine `server_hostname` automatically from the websocket URI.

aaugustin:
Perhaps we should use `server_hostname` if it is provided and fall back to `wsuri.host` if it isn't?

FraMecca:
I think there is no meaningful default.

For example the facebook onion domain has a valid certificate so wsuri.host would be a good guess,
but I have seen in the past invalid ssl certificates on the .onion network.

I think that it is the developer that should provide server_hostname in the same way the custom socket is passed to the parameter of the client.connect.

For this reason I propose to use:
server_hostname = None if not secure else 'localhost' 
on the tests.

aaugustin:
Superseded by #242.

Issue #216, opened 2017-07-23
Rename serve()'s klass to reflect that it can be a function


cjerdonek:
I encountered a situation where it would be somewhat awkward to pass a class to `websockets.serve()` as the `klass` argument and would be more natural to pass a function.  The use case is when subclassing `WebSocketServerProtocol` and wanting to set custom instance attributes to values that are determined at runtime.

This would be similar to (but not quite the same as) the following note in Python's [`AbstractEventLoop.create_connection()`](https://docs.python.org/3/library/asyncio-eventloop.html#asyncio.AbstractEventLoop.create_connection) docs:

> Note: _protocol\_factory_ can be any kind of callable, not necessarily a class. For example, if you want to use a pre-created protocol instance, you can pass `lambda: my_protocol`.

aaugustin:
Sure, go ahead!

cjerdonek:
Will do, thanks! Would you be open to me renaming the argument in the same patch to reflect that klass need not be a class? The main issue I see would be backwards compatibility. Or that can be done later (also see a related minor issue I will be filing shortly ).

aaugustin:
You can rename the argument to `protocol_factory` (good riddance `klass`, you were ugly) and add a backwards-compatibility shim.

That shim would be:
- mentioned in the changelog
- not mentioned in the API docs
- not scheduled for deprecation

aaugustin:
See `legacy_recv` for a very similar case.

cjerdonek:
Thinking more about your comment, do you think `protocol_factory` could cause confusion since it's not the same as the `protocol_factory` passed to [`AbstractEventLoop.create_server()`](https://docs.python.org/3/library/asyncio-eventloop.html#asyncio.AbstractEventLoop.create_server) (and has a different signature)?

In the asyncio docs, `protocol_factory` is called without arguments. But in this case, this "factory" is passed a number of keyword arguments and is used by `serve()` to create a `protocol_factory`.

cjerdonek:
To avoid confusion, `create_protocol` maybe?

aaugustin:
No very strong feelings here. Go for `create_protocol` if you prefer!

cjerdonek:
I posted PR #236 for this.

cjerdonek:
Closed by PR #236.

PR #236, opened 2017-07-29
Address issue #216: rename the klass argument to create_protocol.

cjerdonek:
This addresses issue #216.

One minor note: you'll notice in the signatures that I made the default value `None` instead of the default class. The reason is that this makes the API somewhat friendlier when it comes to composing functions, etc. For example, if an end user defines a wrapper function and wants the same default, they can get the same default by passing `None`. They don't need to hard-code the default value in a second place. Secondarily, when it comes to deciding between `create_protocol` and `klass`, this makes it easier to know which arguments were explicitly passed.

### DIFF
--- a/docs/api.rst
+++ b/docs/api.rst
@@ -32,7 +32,7 @@ Server
 
 .. automodule:: websockets.server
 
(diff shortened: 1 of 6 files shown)


aaugustin:
Excellent! :+1: for using `None` as a default value.

I suggested a few tweaks. If you're OK with them, please merge.

Thanks!

cjerdonek:
Thanks, Aymeric!

Issue #220, opened 2017-07-23
Build wheels on various platforms


aaugustin:
From @messense in https://github.com/aaugustin/websockets/pull/214#issuecomment-317228572:

> For wheel building [cibuildwheel](https://github.com/joerick/cibuildwheel) might be useful if you are willing to do pypi release automatically in CI.

Indeed this appears to be a great way to build wheels on various OSes and perhaps even to upload them automatically to PyPI.

websockets uses a C extension for speedups, not for functionality, making wheels most useful for production systems that don't have a compilation toolchain. I expect websockets to be most commonly deployed on Linux.

wheels for macOS and Windows wouldn't hurt, but I believe these systems are mostly used for development, making it less of a concern. The sdist installs reasonably quickly.

aaugustin:
If we run tox on Windows, we'll hit cross-platform compatibility issues; see the proposal by @mayeut here: https://github.com/aaugustin/websockets/pull/211#discussion_r128901152

PR #239, opened 2017-08-01
Address issue #220: Build wheels on various platforms

mayeut:
This is an attempt (WIP) to build wheels on various platforms using [cibuildwheel](https://github.com/joerick/cibuildwheel) on [travis-ci](https://travis-ci.org) & [appveyor](https://ci.appveyor.com)

So far, wheels are built on all platforms but nothing is done with them.
Results are here:
https://travis-ci.org/mayeut/websockets
https://ci.appveyor.com/project/mayeut/websockets

Known limitations:
- if the C extension build fails, a wheel will be built anyway as was the case for python 3.3 & 3.4 on Windows before 47b95195a28491ad4bfb0925560426bc38e7b8c0 (C extension configured as optional in setup.py)
- tests are not run because of failures
  - `test_close_handshake_timeout` fails on macOS with error `AssertionError: 0.03943201599997792 not less than 0.029 : Too slow: 0.03943201599997792 >= 0.029`
  - `test_server_receives_malformed_request` and `test_server_shuts_down_during_opening_handshake` fail on Windows with error `ConnectionResetError: [WinError 10054] An existing connection was forcibly closed by the remote host`

I can't reproduce those issues (macOS is random, depends on short timings. I don't have a Windows VM available). Full call stacks can be found by following those links: [macOS](https://travis-ci.org/mayeut/websockets/jobs/259532227), [Windows](https://ci.appveyor.com/project/mayeut/websockets/build/3.3.13)

@aaugustin, this is just to get #220 started. Let me know what direction you want this to go to (if any).

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,0 +1,35 @@
+env:
+  global:
+#   Don't attempt to build Python 2.7 wheels; websockets only works on Python 3.
+    - CIBW_SKIP=cp27*
+#   Commented out because tests don't pass reliably on macOS, see #241.
+#    - CIBW_TEST_COMMAND="python3 -m unittest discover websockets"
+
+matrix:
+  include:
+    - dist: trusty
+      sudo: required
+      language: python
+      python: "3.3"
+      services:
+        - docker
+    - os: osx
+      osx_image: xcode8.3
+
+install:
+# Python 3 is needed to run cibuildwheel for websockets.
+  - if [ "${TRAVIS_OS_NAME:-}" == "osx" ]; then
+      brew install python3;
+    fi
+# Install cibuildwheel using pip3 to make sure Python 3 is used.
+  - pip3 install cibuildwheel==0.4.0
+# Create file '.cibuildwheel' so that extension build is not optional (c.f. setup.py).
+  - touch .cibuildwheel
+
+script:
+  - cibuildwheel --output-dir wheelhouse
+# Upload to PyPI on tags
+  - if [ "${TRAVIS_TAG:-}" != "" ]; then
+      python -m pip install twine &&
+      python -m twine upload --skip-existing wheelhouse/*;
+    fi
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -1,0 +1,19 @@
+environment:
+# Don't attempt to build Python 2.7 wheels; websockets only works on Python 3.
+  CIBW_SKIP: cp27*
+# Commented out because tests don't pass reliably on Windows, see #240.
+#  CIBW_TEST_COMMAND: python -m unittest discover websockets
+
+# Since Python 2 is still the default, invoke Python 3 explicitly.
+install:
+  - cmd: C:\Python33-x64\python.exe -m pip install cibuildwheel==0.4.0
+# Create file '.cibuildwheel' so that extension build is not optional (c.f. setup.py).
+  - cmd: touch .cibuildwheel
+build_script:
+  - cmd: C:\Python33-x64\python.exe -m cibuildwheel --output-dir wheelhouse
+# Upload to PyPI on tags
+  - ps: >-
+      if ($env:APPVEYOR_REPO_TAG -eq "true") {
+        Invoke-Expression "python -m pip install twine"
+        Invoke-Expression "python -m twine upload --skip-existing wheelhouse/*.whl"
+      }
--- a/setup.py
+++ b/setup.py
@@ -29,7 +29,7 @@ ext_modules = [
     setuptools.Extension(
         'websockets.speedups',
         sources=['websockets/speedups.c'],
-        optional=True,
+        optional=not os.path.exists(os.path.join(root_dir, '.cibuildwheel')),
     )
 ]
 


aaugustin:
Thanks for working on this! 👍 😀 

Here's my opinion on the limitations.

----

I assume we have to choose between:

1. marking the extension as optional so that `python setup.py install` is able to install the source distribution even if there isn't a working compilation environment
2. failing the build when the extension fails to compile.

Perhaps we can introduce an env var and do something like `optional=not os.environ.get("WEBSOCKETS_C_EXT_REQUIRED")`? (I'll consider a better name...) Even better, perhaps we can detect that we're building a wheel and set `optional=False` in that case?

I don't have very strong opinions there... but I think we can find an acceptable solution without too much work :-)

----

Regarding the test failures, you can try setting the environment variable `WEBSOCKETS_TESTS_TIMEOUT_FACTOR` to `10` or `20` to increase timeouts 10 or 20 times. I'm keeping timeout shorts because I find it a better trade off to have tests complete quickly and the occasional timeout when running tests frequently. However the trade off is different on CI: reliability is very important there.

That said, 0.03943201599997792 looks a lot like 0.029 + 0.010 + something, so that test may actually be broken, perhaps depending on a random race condition :-(

The Windows error looks like an actual bug. It may be fairly easy to fix by someone who has access to a Windows environment.

aaugustin:
You can push 47b95195a28491ad4bfb0925560426bc38e7b8c0 directly to master without waiting for the rest of this PR.

aaugustin:
I took a look at the configurations; I'd like a quick explanation of why they're a bit more elaborate that what the README of cibuildwheel suggests; other than that they look fine.

I'll have to setup TravisCI and AppVeyor with my own account and connect them to this repository.

We can probably stick with the [manuel method](https://github.com/joerick/cibuildwheel#manual-method) for a few releases and then figure out.

mayeut:
Thanks for the feedback.
I pushed 47b95195a28491ad4bfb0925560426bc38e7b8c0 for the MSVC 2010 fix.

I remembered you did not want to clobber the `setup.py` with too much extra code so I kept it simple in a first time with the `optional` limitation. Given how cibuildwheel works, it might not be easy to work with environment variables (not used on linux). I'll look in detecting that a wheel is being built or open an issue/PR at cibuildwheel.

The configuration files are there because all default python environment are `Python 2.7` and cibuildwheel chokes while setting up calling `python setup.py --name`:
```
  File "setup.py", line 11, in <module>
    with open(readme_file, encoding='utf-8') as f:
TypeError: 'encoding' is an invalid keyword argument for this function
```
The extra bits are there to run on Python 3. There's also the `CIBW_SKIP=cp27*` environment variable to skip building Python 2.7 wheels.

mayeut:
Regarding the `WEBSOCKETS_TESTS_TIMEOUT_FACTOR` environment variable, I saw this while working on coverage after pushing this PR.
The issue was on macOS (cibuildwheel keeps environment on macOS) so I could test this a bit. Final results: no more errors with timeout but now `test_remote_close_race_with_failing_connection`is failing with:
```
  File "/Library/Frameworks/Python.framework/Versions/3.4/lib/python3.4/site-packages/websockets/test_protocol.py", line 711, in test_remote_close_race_with_failing_connection
    self.assertConnectionClosed(1011, '')
  File "/Library/Frameworks/Python.framework/Versions/3.4/lib/python3.4/site-packages/websockets/test_protocol.py", line 237, in assertConnectionClosed
    self.assertEqual(self.protocol.close_code, code)
AssertionError: 1000 != 1011
```

aaugustin:
Good. I would find it nice to embed these explanations in the config files, for example like this:

```yaml
environment:
  # Don't attempt to build Python 2.7 wheels; websockets only works on Python 3.
  CIBW_SKIP: cp27*
  # Commented out because tests don't pass reliably on Windows, see #240.
  # CIBW_TEST_COMMAND: python -m unittest discover websockets

# Since Python 2 is still the default, invoke Python 3 explicitly.
install:
- cmd: C:\Python33-x64\python.exe -m pip install cibuildwheel==0.4.0
build_script:
- cmd: C:\Python33-x64\python.exe -m cibuildwheel --output-dir wheelhouse
artifacts:
- path: wheelhouse\*.whl
  name: Wheel
```

aaugustin:
I filed separate issues for the test fixes.

In order to move this forwards without waiting for these fixes, but still run tests before building wheels, we could append `|| true` to the test commands.

aaugustin:
I'd like to make a release fairly soon to see how the C module works in the wild.

@mayeut Let's merge the simplest thing that could possibly work, if you're OK with that. Let me know when you're happy with the state of this PR and I'll try to push a release :-)

cjerdonek:
Do you think you could look at PR #202 before you make a release? My concern is if that PR would change to become API-breaking once the release is made, so it would be no longer possible after te release. (The changes are very slight, but they are still present.) Up to this point, there hasn't been a release with the current `get_response_status()` API. It was originally added for PR #154.

aaugustin:
@cjerdonek Thank for catching this. Indeed, I realize I did a poor job in #154 and I can do better. See #248.

mayeut:
I worked a bit on that:
- Added comments in YML files as requested
- Make the build of the extension mandatory through the presence of a file. Can't be done with environment variable for now c.f. joerick/cibuildwheel#16

Remaining:
- Run tests. We can't append `|| true` to tests commands c.f. joerick/cibuildwheel#7
- Where to upload travis-ci artifacts ?

A bit more on that last point which must be solved before release I think:
Appveyor allows to store artifacts at the end of a build and download them manually. This is not the case for travis-ci which requires to upload artifacts somewhere else c.f. [travis-ci docs](https://docs.travis-ci.com/user/deployment/)

I see a couple easy options for this:
1. deploy automatically to PyPI on tags
2. deploy automatically to GitHub on tags

aaugustin:
Thanks. I think this PR is good to merge as a first step.

For uploading, in fact, the best solution may be to go straight to automatic upload: https://github.com/joerick/cibuildwheel-autopypi-example.

It would be best to test it first by creating a 'websockets-test-do-not-use' package on PyPI, changing the `name` accordingly in `setup.py`, pushing tags and seeing what happens. Once that's done, delete the test package on PyPI and the test tags.

As far as I understand, when merging to master, I will have to:

- configure Travis CI and AppVeyor for aaugustin/websockets
- put my PyPI password in the Travis CI and AppVeyor settings
    - for extra security, I'm planning to create a dedicated PyPI account with permissions restricted to websockets, not my other packages

aaugustin:
BTW I could merge this PR in its current state and set up Travis CI and AppVeyor while we figure out upload. Just let me know if you'd like me do to it now. Thanks!

mayeut:
So, I went straight for automatic upload.
The exemple from cibuildwheel leads to errors at the end of build on appveyor (all was uploaded) so I changed this part of the exemple.
All is working: https://testpypi.python.org/pypi?:action=display&name=websockets-test-upload&version=3.4.1

Your understanding of what you have to do with travis-ci & appveyor configuration is correct.
Here are the environment variables I added to both those services for websockets:

- `TWINE_USERNAME`: This could go in `.travis.yml`and `appveyor.yml`
- `TWINE_PASSWORD`
- `TWINE_REPOSITORY_URL`: needed only if you want to use an alternative URL for upload (as I did with PyPI test server: https://test.pypi.org/legacy/)

aaugustin:
Here's what I did:

- configure Travis CI and AppVeyor, add TWINE_USERNAME and TWINE_PASSWORD environment variables
- bump version number, commit, tag
- `rm -rf dist && python setup.py sdist && twine upload dist/*`
- `git push && git push --tags`

Wheels are arriving here: https://pypi.python.org/pypi/websockets/3.4

- Linux : OK : https://travis-ci.org/aaugustin/websockets/jobs/266488683
- macOS : OK : https://travis-ci.org/aaugustin/websockets/jobs/266488684
    - except there's no wheel for Python 3.3, but I don't care, it will be EOL in September
- Windows : OK : https://ci.appveyor.com/project/aaugustin/websockets/build/1.0.4
    - I had to relaunch the build because I hadn't set credentials correctly at first
    - There's an error in the log but it seems harmless: https://ci.appveyor.com/project/aaugustin/websockets/build/1.0.4#L114

aaugustin:
@mayeut I think we're done here; let me know if I missed anything.

Thanks again for figuring out the details. It was super easy to deploy for me!

mayeut:
Nice!

Regarding your 2 bullet points:

- macOS: no wheel for Python 3.3, this is a limitation of cibuildwheel. I haven't checked why but given your comment. I won't look into this.
- Windows: There's an error in the log. I didn't figure out this one since it's harmless. I don't know much about PowerShell. I would guess twine writes to stderr and PowerShell interprets this.

aaugustin:
Great. Thanks again!

PR #248, opened 2017-08-12
Add process_request() to override request handling.

### LINKED ISSUE #116: WebSockets and HTTP sharing the same port using request headers

aaugustin:
This replaces the get_response_status() API which never made it into a
release (so there's no backwards incompatibility).

Remove a test that depends on get_response_status() being called after
check_request(). The extension point must be before check_request() so
it can handle regular HTTP requests.

Fix #116.

Supersedes #202 #154, #137.

### DIFF
--- a/docs/api.rst
+++ b/docs/api.rst
@@ -42,8 +42,8 @@ Server
    .. autoclass:: WebSocketServerProtocol(ws_handler, ws_server, *, host=None, port=None, secure=None, timeout=10, max_size=2 ** 20, max_queue=2 ** 5, read_limit=2 ** 16, write_limit=2 ** 16, loop=None, origins=None, subprotocols=None, extra_headers=None)
 
         .. automethod:: handshake(origins=None, subprotocols=None, extra_headers=None)
+        .. automethod:: process_request(path, request_headers)
         .. automethod:: select_subprotocol(client_protos, server_protos)
-        .. automethod:: get_response_status()
 
 Client
 ......
--- a/docs/changelog.rst
+++ b/docs/changelog.rst
@@ -14,8 +14,8 @@ Changelog
 * :func:`~websockets.server.serve` can be used as an asynchronous context
   manager on Python ≥ 3.5.
 
-* Added support for rejecting incoming connections by customizing
-  :meth:`~websockets.server.WebSocketServerProtocol.get_response_status()`.
+* Added support for customizing handling of incoming connections with
+  :meth:`~websockets.server.WebSocketServerProtocol.process_request()`.
 
 * Made read and write buffer sizes configurable.
 
--- a/websockets/client.py
+++ b/websockets/client.py
@@ -30,7 +30,7 @@ class WebSocketClientProtocol(WebSocketCommonProtocol):
     @asyncio.coroutine
     def write_http_request(self, path, headers):
         """
-        Write status line and headers to the HTTP request.
+        Write request line and headers to the HTTP request.
 
         """
         self.path = path
@@ -52,7 +52,11 @@ class WebSocketClientProtocol(WebSocketCommonProtocol):
         Read status line and headers from the HTTP response.
 
         Raise :exc:`~websockets.exceptions.InvalidMessage` if the HTTP message
-        is malformed or isn't a HTTP/1.1 GET request.
+        is malformed or isn't an HTTP/1.1 GET request.
+
+        Don't attempt to read the response body because WebSocket handshake
+        responses don't have one. If the response contains a body, it may be
+        read from ``self.reader`` after this coroutine returns.
 
         """
         try:
--- a/websockets/compatibility.py
+++ b/websockets/compatibility.py
@@ -12,6 +12,7 @@ try:                                                # pragma: no cover
                                                     # Python ≥ 3.5
     SWITCHING_PROTOCOLS = http.HTTPStatus.SWITCHING_PROTOCOLS
     # Used only in tests.
+    OK = http.HTTPStatus.OK
     UNAUTHORIZED = http.HTTPStatus.UNAUTHORIZED
     FORBIDDEN = http.HTTPStatus.FORBIDDEN
 except AttributeError:                              # pragma: no cover
@@ -19,6 +20,10 @@ except AttributeError:                              # pragma: no cover
     class SWITCHING_PROTOCOLS:
         value = 101
         phrase = "Switching Protocols"
+
+    class OK:
+        value = 200
+        phrase = "OK"
 
     class UNAUTHORIZED:
         value = 401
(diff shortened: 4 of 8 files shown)


aaugustin:
It would be nice to extract this test: https://github.com/aaugustin/websockets/pull/202/files#diff-7588b4fc16ad7f464574d6147945e27aR313 which exercises a realistic use case for this feature.

aaugustin:
A few comments on the design decisions here...

Historically websockets uses `get_header` and `set_header` functions internally (which are closures over the request and response headers respectively). I'm starting to think this may not be a very wise choice, for two reasons:

1. It doesn't solve the "multi-value dict" problem: these functions only give you one value while multiple values may be provided for a header (which could cause security issues, by ignoring the other values!)
2. It requires creating functions on the fly at runtime, which may be slow (I didn't benchmark).

The new `process_request` API takes headers as a `HTTPMessage` -- which we've settled on for representing multi-value dicts -- and returns lists of pairs -- which are most straightforward to build. The assymetry is a bit annoying but I don't think there's much value in forcing the output to be a `HTTPMessage`.

aaugustin:
@cjerdonek Do you think this is a good way to address the various related issues you've been working on?

aaugustin:
One last thing -- I've toyed with the idea of adding a symmetrical `process_response` hook, but the API is messy to define, because it would have to provide the ability:
    - to access response headers and to modify them
    - to establish the websocket connection, perhaps with modified headers, or to abort it
Also I'm failing to come up with a use case. So I removed that from the patch.

aaugustin:
Hmm, currently this doesn't provide a way to send a response body...

aaugustin:
And now it does!

cjerdonek:
Thanks so much for working to get this API better before releasing! It is an important one.

A few comments:

First, in `WebSocketServerProtocol.handler()`, exceptions in the call to `self.handshake()` are handled by calling `self.writer.write()`. However, `handshake()` itself also includes calls to `self.writer.write()`, which means `handler()` may be asked to write after some writes have already taken place. Thus, it seems like there can be a better separation of responsibilities here. For example, instead of `handshake()` doing any writing, it could return what to write. This would make the exception handling clean, because writes would take place in either the except block, or in the "else" condition (I'm using "else" in the informal sense here), but not both.

Second, it seems a bit klunky / arbitrary to me for `handshake()` to return `path` and to use that as the basis for whether to return early. For one, `path` is already available as `self.path`. It could just as well be a boolean. In terms of my suggestion above, it seems like a more natural solution would be for `handshake()` to return `early_response` and use whether `early_response` is None or non-None as the basis for whether to return early (and acting on `early_response` if non-None).

Lastly, the body of `handshake()` calls `read_http_request()` prior to calling `process_request()`, but `read_http_request()`'s docstring says, "This coroutine assumes that there is no request body." Would the `process_request()` API support accessing the request body? It seems like that would be an important use case. In other words, it seems like `process_request()` shouldn't have to make any assumptions about the form of the request. It should be possible to fully delegate (or override a default delegation).

aaugustin:
> First, in WebSocketServerProtocol.handler(), exceptions in the call to self.handshake() are handled by calling self.writer.write(). However, handshake() itself also includes calls to self.writer.write(), which means handler() may be asked to write after some writes have already taken place. Thus, it seems like there can be a better separation of responsibilities here. For example, instead of handshake() doing any writing, it could return what to write. This would make the exception handling clean, because writes would take place in either the except block, or in the "else" condition (I'm using "else" in the informal sense here), but not both.

> Second, it seems a bit klunky / arbitrary to me for handshake() to return path and to use that as the basis for whether to return early. For one, path is already available as self.path. It could just as well be a boolean. In terms of my suggestion above, it seems like a more natural solution would be for handshake() to return early_response and use whether early_response is None or non-None as the basis for whether to return early (and acting on early_response if non-None).

The documented API of `handshake` has always been that it returns `path`.

Originally `websockets` had a functional design: information was passed through function parameters and return values. Later I stored a lot of information in instance attributes to facilitate extension by subclassing. This creates some inconsistencies and tension in the design.

I tried refactoring `handler()` and `handshake()` to move all I/O outside of `handshake()` in #251 but I don't like the assymetry between the client and the server, and the exception handling in `handler()` was still quite messy.

I'll see if I can do better...

----
> Lastly, the body of handshake() calls read_http_request() prior to calling process_request(), but read_http_request()'s docstring says, "This coroutine assumes that there is no request body." Would the process_request() API support accessing the request body? It seems like that would be an important use case. In other words, it seems like process_request() shouldn't have to make any assumptions about the form of the request. It should be possible to fully delegate (or override a default delegation).

If a user needs access to the request body, they can keep reading from `self.reader` after `read_http_request` has read the headers. I'm updating the docstrings to explain this.

It's up to them to figure out how much to read, based on Connection, Content-Length, etc. or reaching EOF. I still don't intend to write a full HTTP framework —  just the minimum needed to support websockets and reasonable deployment scenarios ;-)

Currently `read_request` is hardcoded to expect a GET HTTP/1.1 request. Request bodies in GET requests are legal but uncommon. Per https://tools.ietf.org/html/rfc7231#section-4.3.1:
> A payload within a GET request message has no defined semantics;
> sending a payload body on a GET request might cause some existing
> implementations to reject the request.

I expect the restriction to HTTP/1.1 to more problematic than the restriction to GET because some HTTP load balancers could make HTTP/1.0 health check requests. We'll see if someone files a ticket about that... Hopefully load balancers that are able to deal with websocket connections efficiently support HTTP/1.1 for health checks...

aaugustin:
I added a commit that improves the situation without making backwards incompatible changes.

Before:

- `handshake()` could write a response in two places (early response, regular handshake response)
- `handshake()` could return three kind of results (None, a path, an exception)

After:

- `handshake()` writes a response if and only if it's a successful handshake response and returns the path in that case (like it used to)
- `handshake()` raises an exception in all other cases

The "if and only if" holds under the assumption that the following four lines can't raise an exception, which seems reasonable:

```python
        assert self.state == CONNECTING
        self.state = OPEN
        self.opening_handshake.set_result(True)

        return path
```

This addresses fully @cjerdonek's second point above by replacing the `return None` case by `raise AbortHandshake` — which is handled together with other exceptions, with minimal changes.

I doesn't address fully his first point. While HTTP responses may still be written by `handshake()` or by `handler()`, I think the separation of responsibilities is now clear enough and there's no risk of writing two responses.

We can still make the backwards-incompatible change I tried in #251 later without affecting the `process_request` API, which is the API we're trying to get right in this PR.

PR #251, opened 2017-08-20
Refactor handshake take 3

aaugustin:


aaugustin:
This is an alternative to #248, not intended to be merged.

Issue #245, opened 2017-08-09
Improve README


aaugustin:
It doesn't quite reflect the amount of effort put into the library. It could use some marketing copy.

Issue #222, opened 2017-07-23
Measure coverage of C code


aaugustin:
@mayeut has been working on this: see https://github.com/aaugustin/websockets/pull/211#issuecomment-317249971 for details and recover the corresponding bits from that PR.

mayeut:
There are multiple options for this:

1. Same as #211 for `tox.ini`. It lacks validation with a `--fail-under` like option. I simplified the one from #211 for Unix only in mayeut/websockets@9163a4786fa7e693a1bd2511a3e4356e70081d1c
2. Use CI to collect coverage information & validate PRs. I implemented this using [travis-ci](https://travis-ci.org/mayeut/websockets/builds/260170245) & [codecov](https://codecov.io/gh/mayeut/websockets/tree/coverage/websockets) in mayeut/websockets@55f5a67653479eabea88d36ce1aad554cfb7e29e and mayeut/websockets@ea7491dc2e48eaec07d556c7a530adb7d6f4211d (this second commit is to get 100% coverage on C extension)
3. A combination of both
4. Other ?

aaugustin:
Right now I'm using Circle CI for running tox on PRs. I recently discovered that it only works for PRs from branches committed to this repo, not from branches in forks. I'm pretty sure that's a matter of configuration; I haven't looked into it.

I switched from Travis CI to Circle CI because I have generally had a very poor experience with Travis CI, mainly with random failures. Circle CI is easier to setup and more reliable in my experience. This isn't a deal breaker, but even if we build wheels on Travis CI (because that's what cibuildwheel supports) I'd rather keep Circle CI as our primary CI service.

I need to think a bit about CodeCov. It seems a bit annoying to depend on CI + a third-party service to validate coverage. Of course, if the alternative is "not validating coverage at all", it's still an improvement...

aaugustin:
Let's add CodeCov.

It's fine if only Python code is covered. I understand what the `bytesnullhook` in https://github.com/mayeut/websockets/commit/ea7491dc2e48eaec07d556c7a530adb7d6f4211d does and I don't want to maintain that in this repository.

aaugustin:
(CodeCov is a bit orthogonal to the original issue here.)

mayeut:
[...]

PR #252, opened 2017-08-20
Add PyPI and CircleCI badges to README

mayeut:
This reflects at first glance that this package is available on PyPI and that CI is in place.

I think that adding TravisCI and AppVeyor badges is not yet something that should be done. Not until wheels can be tested on those platforms.

In my opinion, a codecov badge wouldn't hurt as well c.f. #222. With that, at first glance, people will know package is available on PyPI, tested by CI and that coverage is 100%.

This, I think, is part of marketing tools (people knowing the project already know all of this).

Relates to #245

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -1,5 +1,5 @@
-WebSockets
-==========
+WebSockets |pypi| |circleci|
+============================
 
 ``websockets`` is a library for developing WebSocket servers_ and clients_ in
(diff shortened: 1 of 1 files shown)


aaugustin:
I hate badges. Really.

But, yeah, this library could use more marketing. Let's do this.

PR #254, opened 2017-08-21
Add C coverage using Makefile

mayeut:
This adds C coverage on Unix using the existing `Makefile`.

`circleci.yml` has been updated to run `gcov` file generation for upload in `codecov`.

Relates to #222 

There's another branch that does only part of this in `tox` (only one code path tested) c.f. mayeut/websockets@613eaed0113eeafa9643e329851c7fce446f4844. I feel that it's easier and more readable to do all code path tests in the `Makefile` here. Furthermore, we could add a check on coverage using a simple `grep` on `gcovr` output (requires `gcovr` to be added to `pip install` list in `circleci.yml`)

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,5 +1,6 @@
 *.pyc
 *.so
+*.gcov
 .coverage
 .tox
 build/
--- a/Makefile
+++ b/Makefile
@@ -9,7 +9,19 @@ coverage:
 	python -m coverage run --branch --source=websockets -m unittest
 	python -m coverage html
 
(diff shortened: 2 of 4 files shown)


aaugustin:
My latest thoughts on this...

* The code changes required to get to 100% code coverage make it marginally worse. If we have more C code in the future, I want to be able to return early if something fails. I'd rather mark the error handling paths as untested if we don't have a simple enough way to test them.

* The Makefile is still a bit more involved than I wish it were but I don't see any candidates for optimization. We need a bit more glue for C than for Python, where we can simply run `python -m coverage`.

Issue #240, opened 2017-08-03
Failing tests on Windows


aaugustin:
From #239.

`test_server_receives_malformed_request` and `test_server_shuts_down_during_opening_handshake` fail on Windows with error `ConnectionResetError: [WinError 10054] An existing connection was forcibly closed by the remote host`.

Full stack trace: https://ci.appveyor.com/project/mayeut/websockets/build/3.3.13

aaugustin:
#267 may affect this. It could make things better or worse.

aaugustin:
Apparently #267 fixed it.

Issue #112, opened 2016-05-19
Problem detecting network loss


RouquinBlanc:
Hi Aymeric,

I'm having difficulties detecting a loss of connectivity on the websocket server. Basically, I would like my websocket to timeout quickly when we lose connectivity to the client.

my server side ws_handler is pretty simple, looking about like this:

``` python3
while True:
    data = # Get data from somewhere
    yield from websocket.send(data)
```

When the send buffer will be full, send() becomes a blocking call. If I disconnect the network cable from the client (reading on the websocket), I get quickly to a full send buffer, while TCP retries sending for ages.

I then modified the send to add a timeout myself:

``` python3
while True:
    data = # Get data from somewhere
    try:
        yield from asyncio.wait_for(websocket.send(data), TIMEOUT, loop=loop)
    except asyncio.TimeoutError:
         yield from websocket.close(reason='websocket_timeout')
```

But then the close remains stuck because it does not manage to send the OP_CLOSE frame (blocked on drain()).
I could also wrap the close() into a wait_for() too, but then the websocket would not be effectively closed. I have another coroutine waiting on the websocket.recv() part, and it would be good that it fails on a ConnectionClosed error.

The only way I manage to close the websocket without ending up blocked until TCP connection times out is by doing:

```
websocket.writer._transport.abort()
```

Which does not seem very great: I don't get a chance to set a code and a reason by doing so.

Could you advise me on the best way to achieve closing the connection, without waiting (or at least not too much), and if possible setting properly a code and reason?

Thanks
Jonathan

aaugustin:
I believe this happens because disconnecting the network cable doesn't result in the TCP connection being closed.

Can you check if [connection_lost](https://github.com/aaugustin/websockets/blob/817a14fd3b85c0bdc4449bfe792098eaa2a45bfe/websockets/protocol.py#L647) gets called?

If yes, this looks like a bug.

If not, websockets has no way to know that the remote endpoint went away, so there isn't much I can do.

RouquinBlanc:
connection_lost is not called, probably because the socket has not timed out yet (I guess?).

But:
- isn't there a timeout on the websocket itself, at higher level compared to TCP socket timeout?
- I can stick with wrapping the websocket.send(...) call into a wait_for(), but then when it times out, I want to close properly my websocket, with a proper reason, which currently I can't because the close() will get blocked itself. Do you provide a kind of websocket.force_close(reason='because I decided'), similar to aborting a transport?

[...]

PR #265, opened 2017-09-04
Document the design of this library.

aaugustin:


aaugustin:
@cjerdonek I'm planning to explain how I designed the library. I hope you'll find this useful. We should discuss the section about cancellation after I write it (we disagreed on that yesterday).

aaugustin:
This PR builds on top of #267; only the last three commits really matter.

@cjerdonek: I'd love to know if you find my explanation of the design (a) understandable (b) helpful.

I think it's an important step for enabling other people than myself to hack on the internals!

docs/design.rst
```diff
@@ -0,0 +1,403 @@
...
+Connection termination
+----------------------
+
+:attr:`~protocol.WebSocketCommonProtocol.close_connection_task`, which is
+started when the WebSocket connection is established, is responsible for
+eventually closing the TCP connection.
+
+First :attr:`~protocol.WebSocketCommonProtocol.close_connection_task` waits
+for :attr:`~protocol.WebSocketCommonProtocol.transfer_data_task` to terminate,
+which may happen as a result of:
```
cjerdonek: I haven't reviewed all of PR #267, but I'll stop here because this seems central to the PR.

It seems like it is unnecessarily complicating things to create `close_connection_task` in advance to wait for `transfer_data_task` to terminate. Why not just use the normal control flow of a `try` block that awaits on `transfer_data_task`? That seems like the more natural and obvious thing to do.

Also, the fact that `close_connection()` needs to have an `after_handshake` boolean flag to track where in the code the method is being called seems to support this. It seems like it is compensating for what could be handled more naturally by the normal control flow of `try` blocks.

docs/design.rst
```diff
@@ -0,0 +1,403 @@
...
+- a protocol error, including connection errors: depending on the exception,
+  :attr:`~protocol.WebSocketCommonProtocol.transfer_data_task` fails the
+  WebSocket connection with a suitable code and exits.
+
+:attr:`~protocol.WebSocketCommonProtocol.close_connection_task` is separate
+from :attr:`~protocol.WebSocketCommonProtocol.transfer_data_task` because that
+makes it easier to implement the timeout on the closing handshake. Cancelling
+:attr:`~protocol.WebSocketCommonProtocol.transfer_data_task` creates no risk
+of cancelling :attr:`~protocol.WebSocketCommonProtocol.close_connection_task`
+and thus failing to close the TCP connection.
```
cjerdonek: For these two points, why can't you just pass the timeout when you start closing the handshake? And for cancelling the data transfer task, you can still shield or handle `CancelledError` for closing to happen correctly. IIRC, this seems similar to the structure you had before, but I don't see why you couldn't incorporate the logic you needed.

docs/design.rst
aaugustin: Yes, you're correct, this is the crux of the PR.

The reason for having these two tasks is the following. I want to be able to cancel `transfer_data_task` — which is how the closing handshake times out — without risking cancelling `close_connection_task`.

If I did:
```
try:
    # transfer data
finally:
    # close connection
```
it would be more difficult to guarantee that cancellation can happen in the transfer data block but not in the close connection block.

I start `close_connection_task` as soon as the opening handshake completes rather than when `transfer_data_task` exits so I can await `close_connection_task` without worrying whether it exists.

The `after_handshake` is related to the opening handshake; it allows reusing the same code for closing the TCP connection if something fails during the TCP handshake (that's the `after_handshake=False` path). I tried creating two functions and calling the appropriate one depending on whether the opening handshake has succeeded or not. To me no version looks significantly better than the other. If you have a preference for two separate functions, I'm happy to oblige :-)

docs/design.rst
aaugustin: It was very hard to convince oneself that the previous implementation was robust in the face of an arbitrary sequence of the following events:

- calling `close()`, possibly multiple times (that shouldn't raise an error)
- receiving a close frame from the other end, which initiates the closing handshake
- receiving a TCP half-close or a TCP close from the other end

#112 is perhaps the best example: it's very hard to figure out what happens in such cases.

The new design has much simpler invariants: there's a well defined and easily auditable contract for the two tasks:
- when they run
- when they can be cancelled
- whether they can exit with an exception (no)

Since the two tasks essentially run in sequence — `close_connection_task` waits for `transfer_data_task` to terminate before doing anything — it must be possible to:
- concatenate them
- complete a future at the junction
- await that future instead of awaiting the first task
- checking the value of that future whenever cancellation may propagate the task and let it propagate only if the futur isn't completed yet

I think that would make the code more complex and that's why I stuck with two tasks.

(I'm wondering how much memory overhead this creates. I'm planning to check that. But it won't be much compared to buffers and compression state.)

docs/design.rst
cjerdonek: You already merged the change, but I still don't see why all this complexity is needed. There are *lots* of options to consider within the "one task at a time" approach. It is the idiomatic approach. Have you tried them or discussed them if you needed help?

> it would be more difficult to guarantee that cancellation can happen in the transfer data block but not in the close connection block.

As I said above, there are many options, but why not just expose an API that lets you cancel only the transfer data block and not allow more general cancellation?

Alternatively, you could shield the close connection block from cancellation and/or provide a different way to cancel the close connection block if you want to support canceling that.

> I start close_connection_task as soon as the opening handshake completes rather than when transfer_data_task exits so I can await close_connection_task without worrying whether it exists.

But the normal asyncio approach is simply to await a task when you're creating it, by proceeding sequentially through the code. This problem only exists when you set up the approach of creating tasks in advance before needing them.

aaugustin: Yes, I merged because I wanted to make further improvements to the documentation without creating conflicts — I'm working on them right now.

The main reason why I ended up with the two-task design is that it feels _less complex_ to me.

By separating the TCP layer and the WebSocket layer, I was able to reason about what is expected and what isn't — especially exceptions and cancellation — and write it down (see docs/design.rst). Only a handful of things can happen to each task. As a consequence I'm much more confident in the general logic.

I have to admit that I never trusted in the robustness of the previous implementation (which was the third or fourth iteration already) when it came to closing connections. I'm unsure how concurrent calls to close() or fail_connection() worked. There's a long story of changes and refactorings in this area. I never had a clear mental model of what guaranteed connections would be closed.

That's why #112 remained open for so long and why I was able to fix it today.

Also supporting this change is the fact that it fixed #240 as a side-effect.

That said, I'm not opposed to moving back to a single task model if we manage to maintain the same guarantees of correctness. Perhaps it will be easier to write a provably correct single task implementation after understanding the constraints with the simpler two-task implementation. Seeing the corresponding changes in design.rst will help us double-check the reasoning.

aaugustin: There's exactly four references to `close_connection_task` in the code base. If you'd like to merge it back into `transfer_data_task` while maintaining the same behavior, go for it! We'll merge it if it doesn't look more complex or more difficult to check for correctness.

cjerdonek: Also, I don't know how this is currently addressed, but I think it's important to have a clear understanding of the _API_ you want to support and expose around things like cancellation and closing connections before doing refactoring. Is this currently documented? If you're clear on the API you want and what should be supported, then discussing and reasoning about the implementation is simpler.

The API should also be documented separately from the implementation details. Otherwise, it's not clear what's what.

I would also hope that the two tasks are an implementation detail and not part of the public API. Otherwise, it won't be changeable without breaking things for the end-user.

aaugustin: The public API I want to support is what shows up on this page — mainly from docstrings: http://websockets.readthedocs.io/en/stable/api.html

The policy wrt. cancellation isn't sufficiently spelt out, that's a good point. I think the natural expectation is "cancelling any coroutine raises CancelledError and websockets doesn't break apart". You pointed out earlier the close() swallowed CancelledError; I fixed that :-)

Then there are things "don't leak TCP connections that could take a couple minutes to time out, if not more." which aren't usually in API docs but which I care a lot about.

aaugustin: Cf. #278.

cjerdonek: > I think the natural expectation is "cancelling any coroutine raises CancelledError and websockets doesn't break apart".

But there is a little more to it because you also need to specify behaviorally what occurs when the CancelledError is raised (e.g. does cleanup also occur?). I think the edge cases that might be harder / take more work to document are things like coroutines that call more than coroutine inside in sequence which has already been mentioned (e.g. both transferring data and closing the connection). What do you want to happen there? Maybe there is a general policy that can be used to describe the expectation across the library, or maybe it needs to be documented per function in the docstring (though it should be without referencing internal implementation details).

PR #266, opened 2017-09-10
Allow customizing User-Agent and Server headers, fix #262

### LINKED ISSUE #262: Allow setting user-agent
At the moment it seems like user-agent is being set in handshake(). It would be really useful if a user of the library could set its own user-agent string.

Two ideas I have:
* Either creating new parameter to the connect function, or 
* Allowing **extra_headers** to override the default user-agent header

I think I can provide PR for either of these. Will you be open to any of these ideas?

Thanks!

m1keil:


websockets/client.py
```diff
@@ -243,7 +244,8 @@ def handshake(self, wsuri, origin=None,
             for name, value in extra_headers:
                 set_header(name, value)
 
-        set_header('User-Agent', USER_AGENT)
+        if not is_header_set('User-Agent'):
```
aaugustin: I believe `if 'User-Agent' in request_headers:` would do the job more efficiently.

https://docs.python.org/3/library/email.compat32-message.html#email.message.Message.__contains__

There's no need to allocate a dict, build it from `request_headers`, call its `keys()` method.

The `set_header` function was a cute API idea that I need to get rid of... ;-)

aaugustin:
Tests look good!

I think the implementation can be simplified a bit, see my comment inline.

Could you also add a line in the changelog like "Allowed ``extra_headers`` to override User-Agent"?

Thanks!

websockets/client.py
m1keil: Is this what we're after?

    from email.message import Message

    m = Message()
    for t in request_headers:
        m.add_header(*t)

    if 'User-Agent' not in m:
         set_header('User-Agent', USER_AGENT)

m1keil: Or should I just make request_headers a Message() in the first place and get rid of the lambdas ?

aaugustin: Oops, sorry, we don't have a Message yet at this point, it's just a list of pairs. Leave it as is.

m1keil: Ok no problem!

m1keil:
Updated the changelog.

PR #273, opened 2017-09-25
add unix socket support

PrettyWood:
At the moment only a TCP server can be made.
This PR allows the user to use a unix socket and create a UNIX Domain socket server.

### DIFF
--- a/websockets/server.py
+++ b/websockets/server.py
@@ -574,7 +574,8 @@ def serve(ws_handler, host=None, port=None, *,
           read_limit=2 ** 16, write_limit=2 ** 16,
           loop=None, legacy_recv=False, klass=None,
           origins=None, extensions=None, subprotocols=None,
-          extra_headers=None, compression='deflate', **kwds):
+          extra_headers=None, compression='deflate',
+          unix_socket=False, **kwds):
     """
     Create, start, and return a :class:`WebSocketServer` object.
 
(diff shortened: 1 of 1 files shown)


aaugustin:
As far as I know, the specification of websockets only describes its usage over TCP and optionally TLS. There are some TCP-specific considerations in the design. This makes me uncomfortable about supporting "websockets over unix sockets".

That said, outside of providing defaults for various arguments, `serve` does very little. It's just a helper for creating an asyncio server with `WebSocketServerProtocol`.

You could reuse `WebSocketServerProtocol`, where most of the value of `websockets` lies, and copy-paste the five lines you modified in the PR to start your server.

PrettyWood:
I see what you mean. I hence decided to follow your advice by creating a new `serve` function ,which uses `create_unix_server`. Thanks

aaugustin:
Thanks, it's good to know that this workaround solves your problem. Could you share your alternative version of serve()?

I understand the use case. It's quite reasonable. Let's keep the pull request open for now.

PrettyWood:
Sure !
```@asyncio.coroutine
def serve(ws_handler, path=None):
    loop = asyncio.get_event_loop()
    ws_server = WebSocketServer(loop)
    factory = lambda: websockets.WebSocketServerProtocol(ws_handler, ws_server, loop=loop)
    server = yield from loop.create_unix_server(factory, path)

    ws_server.wrap(server)

    return ws_server```

aaugustin:
Superseded by #281 which provides a first-class API, similar to `create_unix_server` vs. `create_server`

PR #281, opened 2017-10-01
Add unix_server to create a Unix server.

aaugustin:
Based on #273 by @PrettyWood.


aaugustin:
@PrettyWood Can you confirm this covers your use case in #273?

PrettyWood:
:+1: thanks a bunch !

Issue #287, opened 2017-10-13
Support asyncio.Server's sockets member, or other way to access port


earonesty:

earonesty:
The asyncio.Server class supports the "sockets" list.. .for the list of sockets connected.   It, of course, should be a function.

Anyway, the only way to get access to the sockets, including key things like the port number allocated, is to access the WebSocketServer's server member.

Instead, I think this member should be protected (_server), and a sockets() method added.

```
            # see: https://github.com/aaugustin/websockets/blob/master/websockets/server.py#L504
            # and: https://websockets.readthedocs.io/en/stable/api.html?highlight=websocketserver#websockets.server.WebSocketServer
            # and: https://docs.python.org/3/library/asyncio-eventloop.html#asyncio.Server

```

aaugustin:
It's unclear to me that it's a good idea to access the list of sockets. What do you want to do with them?

earonesty:
I need to know what port you are listening on, see tests.

aaugustin:
Here's what I understand from your comments and your patch.

> The return value of `websockets.serve()` should provide a way to tell which port the server is listening on. This would allow calling it with `port=0` to use any available port.
> Once this feature is available, tests could take advantage of it instead of hardcoding port 8642. This would avoid cascading failures when a test fails and doesn't close the connection correctly.
> It could also be interesting to implement a `sockets` method to better mimic the `asyncio.Server` API.

@earonesty Does this summarize correctly your proposal?

aaugustin:
#301 does what I described above.

aaugustin:
Fixed.

PR #288, opened 2017-10-13
Adding a sockets() method to the server

earonesty:
https://github.com/aaugustin/websockets/issues/287


earonesty:
Exposing sockets is a bad idea... instead expose port only.

PR #292, opened 2017-10-23
Alternate method of exposing port

earonesty:
Instead of exposing the sockets themselves.

https://github.com/aaugustin/websockets/issues/287


earonesty:
I think that failed test has something to do with this:

> For a ClientSession with SSL, the application must wait a short duration before closing:
> 
> ...
> # Wait 250 ms for the underlying SSL connections to close
> loop.run_until_complete(asyncio.sleep(0.250))

aaugustin:
Superseded by #301.

Issue #291, opened 2017-10-18
<class 'websockets.py35.server.Serve'> breaks documentation and API


SlightlyCyborg:
https://websockets.readthedocs.io/en/stable/api.html#websockets.server.serve

This link states that websockets.serve() returns a WebSocketServer, however it returns a Serve wrapper with self.server as the actual WebSocketServer

https://github.com/aaugustin/websockets/blob/master/websockets/py35/server.py#L9

cjerdonek:
It seems (untested) like this would be solved if the `__aenter__()` and `__aexit__()` methods were put directly on `WebSocketServer` instead of creating a new wrapper class.

condemil:
I am gonna check that out

condemil:
I checked the solution from @cjerdonek, it looks like `async with` syntax expects `__aenter__()` and `__aexit__()` on `serve`, not on `WebSocketServer` returned by `serve`.

cjerdonek:
I don’t think that makes sense because serve is  a function, and that’s not
how it currently works. It just needs to be on the object returned by the
function (which can be a WebSocketServer object instead of an object with
type the Server wrapper class).

condemil:
@cjerdonek `serve` is a coroutine and that is why it cannot pass back `WebSocketServer` without being await'ed first.

cjerdonek:
On second thought, it doesn't seem like there's a bug here at all.

When the documentation says `serve()` "returns a `WebSocketServer` object," that implicitly means after awaiting. The OP is getting a `Serve` object because the function isn't being awaited. This isn't unique to the "py35" code path. In Python < 3.5, you won't get a `WebSocketServer` object either if you don't await. You'll get a "serve generator object" (i.e. generator-based coroutine).

Perhaps what can be improved in the docs is to say that `serve()` is a coroutine. This is partly an issue with the Sphinx formatting of coroutines. There's been various discussion about how to format coroutines in Python docs. For example, here's the start of one thread:
https://mail.python.org/pipermail/async-sig/2017-June/000316.html

cjerdonek:
Also, just for comparison, if you look at the [`create_server()`](https://docs.python.org/3/library/asyncio-eventloop.html#asyncio.AbstractEventLoop.create_server) docs (which `serve()` wraps and which the `serve()` docs link to), you'll see that it says `create_server()` "Return[s] a Server object," even though you need to await to get that returned object.

condemil:
[...]

PR #294, opened 2017-10-26
Eliminate the need to wrap serve().

cjerdonek:
See issue #291 for where this came up. [edit: fixed issue-number reference]


cjerdonek:
Another reason for this is to make Python ≥ 3.5 the "normal" case, as opposed to vice versa.

cjerdonek:
I see now that `client.connect()` can be given a similar treatment.

cjerdonek:
@aaugustin If you're open to this refactoring, I'll update this PR for the client side as well.

aaugustin:
I agree that making Python ≥ 3.5 the default case is nice. That was also my first thought when reading the patch.

I also agree that the client needs the same treatment as the server.

I'm tempted to bikeshed some names, for example `asyncio_server = yield from self._server_coro` looks slightly developer-unfriendly to me. (I'm fully expecting many names I picked myself to be equally unfriendly to others...) I'll try to find some time on Wednesday or next weekend and see if I can find something better.

Thanks for the proposal, let's do this.

cjerdonek:
Thanks. Just to summarize the naming issues / constraints I see:

* `server` is currently already being used to refer to the `WebSocketServer` object `ws_server`, by way of being assigned to `Serve.server`. It would probably be better to be calling that attribute `Serve.ws_server` to make clear that the value is not an `asyncio.Server` object. If that were done, then `asyncio_server` could be renamed to `server` and the problem would be solved. However, I was thinking backwards compatibility would prevent that change. Unless I'm not looking in the right spot, `Serve.server` doesn't seem to be documented, so maybe it's not part of the public API after all and so can still be changed?
   
* Re: `self._server_coro`, yes, that name can definitely be better. It represents the coroutine object created from asyncio's `create_unix_server()` or `create_server()` and that yields an `asyncio.Server` object. `create_coro` would be better. Or, if you don't like the `coro` suffix, then maybe something like `creating_server` to distinguish it from the original `create_server` coroutine _function_.

cjerdonek:
Okay, I chose better names and updated client.py with the new pattern.

cjerdonek:
I've added this to 4.0 because (1) it's done, (2) it "may" be considered backwards incompatibility (since `Serve.server` is getting renamed to `Serve.ws_server`), (3) it's not too hard of a change, and (4) it results in a simplification of the code base to build on going forward.

aaugustin:
Yes, it was on my radar, I'm planning to review and merge it today. Thanks!

websockets/client.py
```diff
@@ -272,129 +273,153 @@ def handshake(self, wsuri, origin=None, available_extensions=None,
...
-            extra_headers=protocol.extra_headers,
+    def __init__(self, uri, *,
+                 create_protocol=None,
+                 timeout=10, max_size=2 ** 20, max_queue=2 ** 5,
+                 read_limit=2 ** 16, write_limit=2 ** 16,
+                 loop=None, legacy_recv=False, klass=None,
+                 origin=None, extensions=None, subprotocols=None,
+                 extra_headers=None, compression='deflate', **kwds):
+        """
+        This coroutine connects to a WebSocket server at a given ``uri``.
```
aaugustin: That's slightly misleading because __init__ isn't — and cannot be — a coroutine (at least not until Python gets async object initialization, which I assume is a can of worms).

On Python ≥ 3.5, initializing a `Connect` instance returns an awaitable, so usage is effectively the same as a coroutine.

This problem actually dates back to the introduction to the `Connect` class. It predates this PR.

websockets/client.py
```diff
@@ -272,129 +273,153 @@ def handshake(self, wsuri, origin=None, available_extensions=None,
...
+        else:
+            # If sock is given, host and port mustn't be specified.
+            host, port = None, None
+
+        self._wsuri = wsuri
+        self._origin = origin
+
+        # This is a coroutine object.
+        self._creating_connection = loop.create_connection(
+            factory, host, port, **kwds)
```
aaugustin: I see that the "split point" between `__init__` and `__await__` occurs just before the first `yield from`.

It seems a bit weird to split the logic that way, but since there's no point in calling `__init__` without `__await__`, I guess that's just pragrmatic.

PR #302, opened 2017-11-01
Manage connection states with an enum.

### LINKED ISSUE #284: Make websocket state human-understandable
Currently, when troubleshooting the state of a connection, if you log the "state" attribute, you just see a number like 1. It would be nice if this was something more human-understandable. The states are [here](https://github.com/aaugustin/websockets/blob/5ce75688f59b017c2f0d4efebff8e9022f0e3d67/websockets/protocol.py#L32):

    CONNECTING, OPEN, CLOSING, CLOSED = range(4)

This could be done with an enum.

aaugustin:
Fix #284.

### DIFF
--- a/websockets/protocol.py
+++ b/websockets/protocol.py
@@ -10,6 +10,7 @@ import asyncio
 import asyncio.queues
 import codecs
 import collections
+import enum
 import logging
 import random
 import struct
@@ -29,7 +30,8 @@ logger = logging.getLogger(__name__)
 
 # A WebSocket connection goes through the following four states, in order:
 
-CONNECTING, OPEN, CLOSING, CLOSED = range(4)
+class State(enum.IntEnum):
+    CONNECTING, OPEN, CLOSING, CLOSED = range(4)
 
 # In order to ensure consistency, the code always checks the current value of
 # WebSocketCommonProtocol.state before assigning a new value and never yields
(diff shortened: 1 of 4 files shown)


aaugustin:
@cjerdonek Hit the merge button if you're happy with that!

cjerdonek:
Looks good, thanks! I also like the abbreviated style of defining an enum, which I hadn't seen before.

cjerdonek:
(Was just waiting for the test run to complete.)

websockets/protocol.py
```diff
@@ -217,8 +219,8 @@ def connection_open(self):
 
         """
         # 4.1. The WebSocket Connection is Established.
-        assert self.state == CONNECTING
-        self.state = OPEN
+        assert self.state == State.CONNECTING
```
RemiCardona: The doc says to use `is` to compare enums. It was recently mentioned to me and it didn't really come naturally to me. I guess it sort of makes sense if one wants to force the use of enums and forgo the `int` capability (since `State` is an `IntEnum`).

Anyway, I'm not 100% sold on this whole `is` thing, but I thought I'd mention it and see what you guys think.

aaugustin: That's... unsettling. But why not. Let's make the change if that's the best practice. Perhaps it's faster.

aaugustin: Fixed in f730fe091fdb9aab8b608e710d260efdf4e8a077.

Issue #305, opened 2017-11-02
loop behavior change with 4.0


cjerdonek:
I just want to point out an issue I noticed when trying 4.0. This may very well be an acceptable behavior change, but I at least wanted to point it out here for posterity and so it can be discussed.

Basically, when I tried using `serve()` in previously working code, I got an error like the following:

> RuntimeError: Task <Task pending ...> got Future <_GatheringFuture pending> attached to a different loop

The stack trace implicates this line:

https://github.com/aaugustin/websockets/blob/93c7bc1446f1f9e99b0919dceacf84fd92832811/websockets/server.py#L725

I believe the issue is that whereas before the line [here](https://github.com/aaugustin/websockets/blob/93c7bc1446f1f9e99b0919dceacf84fd92832811/websockets/server.py#L679) in `serve()` getting the `loop` for the server:

    if loop is None:
        loop = asyncio.get_event_loop()

was happening within code that had to be awaited, in 4.0 it now happens in `__init__()` (due to the changes I made yesterday).
 https://github.com/aaugustin/websockets/blob/93c7bc1446f1f9e99b0919dceacf84fd92832811/websockets/server.py#L679

If your code calls `asyncio.new_event_loop()` between when `serve()` is called and when the coroutine object return value is awaited, you will get the error.

I believe the fix for people like me is simply to arrange not to call `serve()` until it is ready to be awaited.

Alternatively, `websockets` can be changed to move the code in `__init__()` to `__await__()` (removing the "split"), but this would require saving all of the args as attributes on the `Serve` class, which would make that code a little annoying.

I do think the code for `>= 3.5` is now easier to reason about because you can see it there in front of you all in one place. However, I'm not yet sure if there is a more elegant way of handling all of it that will satisfy the various constraints (e.g. including issue #300).

PS - by inspection, I believe the same issue is present for the client-side `connect()` as well.

aaugustin:
Thanks for reporting this issue.

> If your code calls asyncio.new_event_loop() between when serve() is called and when the coroutine object return value is awaited, you will get the error.

That's fine. `serve()` implicitly depends on the current event loop when you don't pass an explicit `loop` argument.

If you change the current event loop after calling `serve()`, you can't use the coroutine it returned. This is a reasonable.

Previously, that implicit dependency only manifested when you awaited the coroutine. This was an implementation detail.

This bug report should allow other people who hit this problem to solve it quickly.

cjerdonek:
[...]

Issue #300, opened 2017-11-01
`connect` and `serve` don't report themselves as coroutines


aaugustin:
```python
>>> from websockets.client import connect
>>> import asyncio
>>> asyncio.iscoroutine(connect)
False
```

I believe this is caused by the wrapper class that makes them async context managers.

cjerdonek:
(I think you want to be calling `asyncio.iscoroutine()` on the return value rather, and `asyncio.iscoroutinefunction()` on the callable.)

Technically, what they are though is ["awaitable."](https://www.python.org/dev/peps/pep-0492/#await-expression)

Yes, this stems from the weirdness of having an async context manager class _also_ being awaitable itself. I think this is the only place I've seen something trying to be both at the same time.

aaugustin:
I wasn't sure whether connect or its return value was supposed to be a coroutine, thanks for filling the gap...

That said I tested both before filing this issue and both returned `False` so the issue is still valid — albeit very minor :-)

cjerdonek:
A couple questions:

* Is it actually important that `iscoroutinefunction()` return `True`?
* I also wonder if it's even possible for something to simultaneously be both a true coroutine and an async context manager -- as opposed to "faking" it so that the function returns True even though it's not really a coroutine. Did you have an idea on how to solve this?

aaugustin:
I'm not sure! Investigation needed.

cjerdonek:
Today I asked about this issue on the async-sig list (originally in general terms), to find out if there was any advantage for a function (and/or its return value) having `asyncio.iscoroutinefunction()` (resp. `asyncio.iscoroutine()` return `True`:
https://mail.python.org/pipermail/async-sig/2017-November/000397.html

The answer seems to be that there isn't really an advantage, in which case I think this issue can be closed after updating the documentation.

The docstrings for `serve()` and `connect()` can be changed from saying each is a "coroutine" to saying each is a "function returning an _awaitable_ that in returns a `WebSocketServer` (resp. `WebSocketClientProtocol`) object."

aaugustin:
Great, thanks for getting that information.

I'm hesitating about changing the documentation. On one hand "callable returning an awaitable" describes the implementation more accurately. On the other hand the behavior is indistinguishable from a "coroutine" for practical purposes so perhaps it's best to keep the description simple?

cjerdonek:
[...]

PR #306, opened 2017-11-02
Fix issue #305: move initialization to __await__().

cjerdonek:
This might be a throwaway PR now that @aaugustin said he didn't think issue #305 needed to be fixed.

I had already written the code though for this PR. I thought it might be worth publishing since the pattern it uses could be useful in the future if we consider other ways to refactor `Connect` and `Serve` (e.g. for issue #300).

### DIFF
--- a/websockets/client.py
+++ b/websockets/client.py
@@ -321,13 +321,17 @@ class Connect:
 
     """
 
-    def __init__(self, uri, *,
-                 create_protocol=None,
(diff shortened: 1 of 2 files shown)


cjerdonek:
PS - if we changed our minds and did want to use this PR, I would also want to add a test for the behavior change about awaiting, and perhaps add a code comment, too.

PR #308, opened 2017-11-07
Fixed infinite loop in "Graceful Shutdown" Examples when client closed the connection

antohaby:
Replaced "pass" statement by "break" inside echo handler loop.
If a client closes connection it causes an infinite loop and took 100% CPU.
"websocket.recv()" always throws "websockets.ConnectionClosed" so "pass" statement just
continue loop over and over again.


aaugustin:
Oops!

Issue #309, opened 2017-11-09
TypeError in WebSocketServer.wait_closed() since tasks not initialized


cjerdonek:
If an exception occurs in `WebSocketServerProtocol.process_request()` and the server is closed, `WebSocketServer.wait_closed()` can raise an exception like the following:

      File "/.../websockets/server.py", line 582, in wait_closed
        loop=self.loop)
      File "/usr/local/lib/python3.6/asyncio/tasks.py", line 311, in wait
        fs = {ensure_future(f, loop=loop) for f in set(fs)}
      File "/usr/local/lib/python3.6/asyncio/tasks.py", line 311, in <setcomp>
        fs = {ensure_future(f, loop=loop) for f in set(fs)}
      File "/usr/local/lib/python3.6/asyncio/tasks.py", line 526, in ensure_future
        raise TypeError('An asyncio.Future, a coroutine or an awaitable is '
    TypeError: An asyncio.Future, a coroutine or an awaitable is required

This looks to be because the tasks that the server is waiting on in `wait_closed()` are not yet initialized (since the initialization was interrupted by the exception in `process_request()`):

https://github.com/aaugustin/websockets/blob/136d9bc9a917f7f5fdde4001e801b591ee1b4a0b/websockets/server.py#L578

cjerdonek:
Looking at the code, it looks like it's the `close_connection_task` which is at fault because it isn't initialized until `connection_open()`, which happens at the end of `handshake()` (and in particular after `process_request()`).

cjerdonek:
Actually, it's possible this is an issue also in the case that `process_request()` raises no exception but returns a non-None value. I'll have to look into it more.

cjerdonek:
[This discussion](https://github.com/aaugustin/websockets/pull/265/files#r140941685) on the topic of whether there should be a separate `close_connection_task` (at the time it was introduced) is also relevant.

[...]

PR #310, opened 2017-11-10
Fix issue #309: server.wait_closed()

cjerdonek:
Fixes issue #309.

### DIFF
--- a/websockets/server.py
+++ b/websockets/server.py
@@ -572,13 +572,8 @@ class WebSocketServer:
         """
         # asyncio.wait doesn't accept an empty first argument.
         if self.websockets:
-            # Either the handler or the connection can terminate first,
-            # depending on how the client behaves and the server is
-            # implemented.
             yield from asyncio.wait(
-                [websocket.handler_task for websocket in self.websockets] +
-                [websocket.close_connection_task
-                    for websocket in self.websockets],
+                [websocket.handler_task for websocket in self.websockets],
                 loop=self.loop)
         yield from self.server.wait_closed()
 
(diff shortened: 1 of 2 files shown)


aaugustin:
Still working on #309...

Apart from that, the decorator that allows writing async tests is a great idea! Two suggestions to make it even better:

- It could apply `@asyncio.coroutine` automatically to remove the need to stack two decorators on async tests.
- I'm not completely sold on the `@make_sync` name. `@async_test` could work — decorator names usually describe a feature of the object being decorated. Or perhaps `@run_in_event_loop`.

aaugustin:
Currently each test already run it its own event loop:

```
    def setUp(self):
        self.loop = asyncio.new_event_loop()
        asyncio.set_event_loop(self.loop)

    def tearDown(self):
        self.loop.close()
```

cjerdonek:
Oh, right. It's been several months since I looked. What I may have been thinking of then is that there can be different decorators that each do varying levels of setup (e.g. (1) doing nothing other than running the loop, (2) also starting and stopping the server, (3) also starting and stopping the client _and_ server, etc). And depending on things, perhaps it could also make sense to put starting and stopping the loop in the decorators instead of splitting things by putting some logic in the setUp and tearDown and some in the decorators.

aaugustin:
Yes, that makes sense. Good ideas.

aaugustin:
and I neglected websockets for several months as well, nothing wrong with that ;-)

aaugustin:
I filed #403 to follow up on the async test refactor.

PR #315, opened 2017-11-30
Emit a debug log message when ignoring unsolicited pongs.

neilconway:
An unsolicited pong likely indicates an error or misconfiguration, so it seems worth logging. I chose `debug`; you could make the argument for logging it at a higher log-level.

### DIFF
--- a/websockets/protocol.py
+++ b/websockets/protocol.py
@@ -614,6 +614,8 @@ class WebSocketCommonProtocol(asyncio.StreamReaderProtocol):
                     while ping_id != frame.data:
                         ping_id, pong_waiter = self.pings.popitem(0)
                         pong_waiter.set_result(None)
+                else:
+                    logger.debug("%s ignoring unsolicited pong", self.side)
 
             # 5.6. Data Frames
             else:


aaugustin:
An unsolicited pong is a perfectly legitimate way to keep the TCP connection open:

> A Pong frame MAY be sent unsolicited.  This serves as a
> unidirectional heartbeat.  A response to an unsolicited Pong frame is
> not expected.

-- https://tools.ietf.org/html/rfc6455#section-5.5.3

All messages are already logged at the debug level.

So I don't see a reason to add more logging in that case.

neilconway:
Okay, fair enough. The context in which I came across this was interop'ing with a broken WebSocket implementation that always sends pong messages with an empty payload; it took quite a bit of debugging to understand that the future returned by `websocket.ping()` was never being completed because although a PONG was received, it was silently discarded because the pong's payload didn't match. Maybe that is an unusual situation, though...

aaugustin:
Ah, that's an interesting bug.

Activating logs at the DEBUG level will reveal it but only if you're familiar with the websocket protocol.

aaugustin:
I'm going to add detailed logs about everything websockets does automatically with pings and pongs.

Issue #318, opened 2017-12-07
A Future or coroutine is required


TheDeadOne:
I tried to run a [basic example](https://websockets.readthedocs.io/en/stable/intro.html#basic-example) from the documentation and got an error:

```
Traceback (most recent call last):
  File "hello.py", line 14, in <module>
    asyncio.get_event_loop().run_until_complete(start_server)
  File "c:\python3\Lib\asyncio\base_events.py", line 322, in run_until_complete
    future = tasks.ensure_future(future, loop=self)
  File "c:\python3\Lib\asyncio\tasks.py", line 531, in ensure_future
    raise TypeError('A Future or coroutine is required')
TypeError: A Future or coroutine is required
```

Python 3.5.0
websockets 4.0.1

cjerdonek:
Thanks for reporting the Python version. Can you try with Python 3.5.1 or higher? `asyncio.ensure_future(coro_or_future)` was changed to accept any awaitable only in Python 3.5.1:

https://docs.python.org/3/library/asyncio-task.html#asyncio.ensure_future

(Technically, perhaps this means websockets doesn't support Python 3.5.0.)

cjerdonek:
Maybe the docs section linked in the OP should be changed from "This section assumes Python ≥ 3.5" to "This section assumes Python ≥ 3.5.1."

aaugustin:
Actually we should say that we only support the latest micro release.

This is how Django documents this policy: https://docs.djangoproject.com/en/2.0/faq/install/#what-python-version-can-i-use-with-django

What's good enough for Django is good enough for websockets ;-)

pilotandy:
I can confirm that it work with 3.5.1, but not 3.5.0.  However, in my case, I'm not the one who can upgrade the version of python in our production environment.  Is there something I can do to get it to work for 3.5.0?

aaugustin:
Try using an earlier version of websockets. 3.4 may work; this looks related to changes we made in 4.0.

PR #324, opened 2017-12-15
Use @asyncio.coroutine decorator on versions 3.5.0 and below.

pilotandy:
Changed in reference to: #318 
Tested and works with versions 3.5.0 and version 3.6.0

### DIFF
--- a/websockets/server.py
+++ b/websockets/server.py
@@ -745,7 +745,7 @@ def unix_serve(ws_handler, path, **kwargs):
     return serve(ws_handler, path=path, **kwargs)
 
 
-if sys.version_info[:2] <= (3, 4):                          # pragma: no cover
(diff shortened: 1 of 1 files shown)


aaugustin:
To me this seems like an improvement. @cjerdonek What do you think?

The documentation may need to be update to say 3.5.1 instead of 3.5 in some places.

cjerdonek:
It seems like it can't hurt, though it goes against your idea to support only the latest micro version (which I thought was a good idea). Maybe you can still do that, but also consider making exceptions on a case by case basis.

Also, I would add a code comment explaining the reason for stopping at 3.5.0.

aaugustin:
Thanks for the pull request. The same changes are required in client.py and in the test suite. It also needs comments and documentation. I created #376 with these additional changes.

Issue #317, opened 2017-12-07
TypeError on forced exit


abalkin:
I have an application that uses websockets to connect to a wss server and processes incoming messages.  Occasionally when I stop the application with <kbd>Ctrl</kbd> <kbd>C</kbd>, I get the following traceback:

```
^C2017-12-07 12:55:33,371 INFO Got signal SIGINT: exit
Traceback (most recent call last):
...
    async for msg in ws:
  File "/Users/a/.virtualenvs/3/lib/python3.6/site-packages/websockets/py36/protocol.py", line 15, in __aiter__
    yield await self.recv()
  File "/Users/a/.virtualenvs/3/lib/python3.6/site-packages/websockets/protocol.py", line 323, in recv
    raise ConnectionClosed(self.close_code, self.close_reason)
  File "/Users/a/.virtualenvs/3/lib/python3.6/site-packages/websockets/exceptions.py", line 147, in __init__
    if 3000 <= code < 4000:
TypeError: '<=' not supported between instances of 'int' and 'NoneType'
```

I think `ConnectionClosed` exception constructor should allow `code` to be None.

cjerdonek:
When addressing this issue, in addition to allowing `code` to be None, it would be good to check that websockets is properly setting the code attribute in each place that it can. I'm not sure if the exception in this instance was raised by the user or websockets. @abalkin, do you know?

abalkin:
The application in which I encountered this error has a signal handler that stops the loop and cancels all running tasks.  I believe the error happens when the task that reads from the websocket an puts messages on the message queue gets cancelled before the task that takes the messages form `self.messages` and gives it to the application.  Maybe the code and reason should be set in [transfer_data](https://github.com/aaugustin/websockets/blob/136d9bc9a917f7f5fdde4001e801b591ee1b4a0b/websockets/protocol.py#L502).

earonesty:
I got the same issue.

aaugustin:
Yes, this is a good suggestion. I'm trying to write a test to prevent future regressions.

aaugustin:
`test_recv_on_closing_connection_local` and `test_recv_on_closing_connection_remote` were supposed to test this, however the connection state is already CLOSED instead of CLOSING when they reach the code we're discussing here.

Delaying the response of the closing handshake gives us the expected state but breaks a bunch of other tests. This requires deeper investigation.

PR #327, opened 2017-12-22
TypeError on forced exit

earonesty:
Fixes https://github.com/aaugustin/websockets/issues/317

### DIFF
--- a/websockets/protocol.py
+++ b/websockets/protocol.py
@@ -320,7 +320,9 @@ class WebSocketCommonProtocol(asyncio.StreamReaderProtocol):
         else:
             next_message.cancel()
             if not self.legacy_recv:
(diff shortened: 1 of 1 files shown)


Gorialis:
CI fails on this due to the assertion (unreachable within the test suite) - given that the assert is ignored when Python is run in optimization mode (making it not a very good way to raise a backup exception here anyway), I think it can be safely removed.

aaugustin:
I'd rather put an assertion on the connection state than an assert False, as shown in #378.

Gorialis:
Yeah, looks good to me.

aaugustin:
Fixed in 402059e4a46a764632eba8a669f5b012f173ee7b. Thanks for suggesting this approach.

PR #342, opened 2018-02-05
Add zip_safe=True to setup.py

### LINKED ISSUE #341: `easy_install websockets` on Windows with Python 3.5 fails to build.
I don't know if this is a duplicate, it seems like it would be a common issue, but I couldn't find it while searching.

Using `easy_install` to download and install websockets on Windows 10, in a Python 3.5 VirtualEnv.

This is the output:
```
zip_safe flag not set; analyzing archive contents...
websockets.__pycache__.test_client_server.cpython-35: module references __file__
 File "c:\src\myproject\venv\lib\site-packages\websockets-4.0.1-py3.5-win-amd64.egg\websockets\py36\protoco
l.py", line 15
   yield await self.recv()
   ^
SyntaxError: 'yield' inside async function

zip_safe flag not set; analyzing archive contents...
error: Setup script exited with error: command 'C:\\Program Files (x86)\\Microsoft Visual Studio 14.0\\VC\\BIN\\x86_amd6
4\\cl.exe' failed with exit status 2
```

It seems the websockets-4.0.1-py3.5-win-amd64.egg bundle contains /py36/ directory that does not compile in python3.5.

It does not fail when using PIP, because I guess it downloads the compiled wheels.

ashleysommer:
add `zip_safe=True` to setup.py to tell easy_install and setuptools that the code should not be analyzed before installing.
This avoids the problem caused by evaluating code in the /websockets/py36/ directory when running under Python 3.5.
Fixes #341

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -60,5 +60,5 @@ setuptools.setup(
     packages=packages,
     ext_modules=ext_modules,
     include_package_data=True,
-    zip_safe=True
+    zip_safe=True,
 )
--- a/setup.py
+++ b/setup.py
@@ -60,4 +60,5 @@ setuptools.setup(
     packages=packages,
     ext_modules=ext_modules,
     include_package_data=True,
+    zip_safe=True
 )


aaugustin:
I'm +0 on merging this PR.

The distribution includes tests (I don't have strong feelings about this, some people do). They won't run properly from a zip install because they need a certificate:
```python
testcert = os.path.join(os.path.dirname(__file__), 'test_localhost.pem')\
```
I doubt anyone cares about this problem.

The line added in this PR needs a trailing comma.

cjerdonek:
testcert = os.path.join(os.path.dirname(__file__), 'test_localhost.pem')

Can this be adjusted easily to work with zip installs out of curiosity?

ashleysommer:
@cjerdonek 
>Can this be adjusted easily to work with zip installs out of curiosity?

I ran into the same problem with a package last week, I was trying to have my `requirements.txt` file included in the zip install for the package.

In my case, the answer was to include `requirements.txt` in the `MANIFEST.in` file, that ensures it is installed in the zip package.

It appears that `test_localhost.pem` is _already_ included in the MANIFEST file, so it might not be a problem.

ashleysommer:
I just did a test to confirm the ^above^.

I did `easy_install --zip-ok websockets` to force websockets to install as a zip package.

Then I ran this test script:
```python
from websockets import test_client_server
found_file = test_client_server.testcert
print(found_file)
with open(found_file) as f:
    cert = f.read()
    print(cert)
```

It worked! It was able to find the file inside the egg, the path resolved correctly, and read and printed the contents of the cert just fine.

So the modification to include `test_localhost.pem` in the zip file is not required because it is already in place.

PR #345, opened 2018-02-11
cheatsheet: `ping()` only requires a single `await`

matthiask:


matthiask:
Oh, I'm sorry, I obv. misunderstood something here.

aaugustin:
I know, it's tricky. If that's any comfort, at some point, I published documentation that contained the same mistake you just made.

`await ws.ping()` sends the ping, it's a coroutine because sending a ping does I/O (first await).

It returns a Future that you can await if you want to wait for the corresponding pong (second await).

PR #348, opened 2018-02-17
Remove useless warning when closing a connection

mattpepin:
Websocket can print a useless warning while the connection closes if the other side closes the socket first (ConnectionResetError). This commit removes this warning if the state is CLOSING or CLOSED.

### DIFF
--- a/websockets/exceptions.py
+++ b/websockets/exceptions.py
@@ -144,9 +144,9 @@ class ConnectionClosed(InvalidState):
         self.code = code
         self.reason = reason
         message = "WebSocket connection is closed: "
-        if 3000 <= code < 4000:
+        if code and 3000 <= code < 4000:
             explanation = "registered"
-        elif 4000 <= code < 5000:
+        elif code and 4000 <= code < 5000:
             explanation = "private use"
         else:
             explanation = CLOSE_CODES.get(code, "unknown")
--- a/websockets/protocol.py
+++ b/websockets/protocol.py
@@ -510,7 +510,8 @@ class WebSocketCommonProtocol(asyncio.StreamReaderProtocol):
         except PayloadTooBig:
             yield from self.fail_connection(1009)
         except Exception:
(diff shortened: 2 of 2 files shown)


aaugustin:
I believe it shouldn't be possible to reach this code path when the connection status is CLOSING or CLOSED. Can you show me how this happens?

mattpepin:
In my case, I created a direct port of Puppeteer from JS (node) to Python 3.6 using asyncio (soon to be on GitHub). Puppeteer is Google's tool to control Chrome/Chromium using DevTools via websockets for automated tests. When the tests are done, Puppeteer kills -9 Chrome (no need to ask gently since it uses a temporary profile) and cleans up at the same time (which closes the client websocket). When this happens, the connection is reset while Websocket is in a closing state. I can create a small repro case if you want, without all the bells and whistles.

aaugustin:
Okay, this may produce an unexpected result. I should take a look at this.

My only concern is not to let errors pass silently, which could make it harder to diagnose mistakes in the connection handler.

mattpepin:
I guess I could additionally explicitly check for ConnectionResetError in the if check.

mattpepin:
I get another related error, worse since it's not only a warning but a crash. Sometimes, instead of a ConnectionReset, I do get a connection_lost(), which puts the state = State.CLOSED. Then (almost simulteneously), write_frame(OP_CLOSE, frame_data) fails, since there's an InvalidState check in protocol.py, line 642.

aaugustin:
@mattpepin Can you test if #389 solves the problem which the first commit in your PR addresses?

I don't want silence all exceptions in the CLOSING or CLOSED state because they can contain useful debug information about why the connection died. That's the only way to know when a websocket connection handler raises an exception.

The problem targeted by your second commit was fixed in 402059e4a46a764632eba8a669f5b012f173ee7b.

mattpepin:
Yes, of course. Not sure if I'll have time tomorrow, but I will report back by tuesday!

PR #351, opened 2018-02-20
Update _creating_connection

EvieePy:
Supports 3.7 and maintains current support.

### DIFF
--- a/websockets/client.py
+++ b/websockets/client.py
@@ -5,6 +5,7 @@ The :mod:`websockets.client` module defines a simple WebSocket client API.
 
 import asyncio
 import collections.abc
+import inspect
 import sys
 
 from .exceptions import (
@@ -390,7 +391,10 @@ class Connect:
         yield from self.ws_client.close()
 
     def __await__(self):
-        transport, protocol = yield from self._creating_connection
+        if inspect.isawaitable(self._creating_connection):
+            transport, protocol = yield from self._creating_connection.__await__()
+        else:
+            transport, protocol = yield from self._creating_connection
 
         try:
             yield from protocol.handshake(


cjerdonek:
Out of curiosity, what is the change in Python 3.7 that makes this PR
necessary? Was there a backwards incompatible change in 3.7?

aaugustin:
I have the same question as Chris, I'd like to understand why we can't have a single implementation that works on all Python versions before making this change.

EvieePy:
My apologies I wrote this a day after a 3 week hospital stay. I'm not sure on the best implementation for this.

Obviously `inspect.isawaitable` is not compat... And wrapping it in a try/except and implementing the same `__await__()` behavior seems fairly hacky. Also this is obviously going to occur in other parts of the lib, and since I have no energy of current to investigate it further, this should obviously be closed.

A quick test on 3.7
```py
import asyncio
import websockets


loop = asyncio.get_event_loop()


class TestServer:

    async def _websocket_handler(self, ws, request):
        pass

    async def _start_server(self):
        server = await websockets.serve(self._websocket_handler, host='...', port=...)


loop.create_task(TestServer()._start_server())
loop.run_forever()
```

Produces the following:
```
Task exception was never retrieved
future: <Task finished coro=<TestServer._start_server() done, defined at C:/Users/Mysterial/PycharmProjects/webstests/wstest.py:13> exception=TypeError("cannot 'yield from' a coroutine object in a non-coroutine generator")>
  File "C:\Users\Mysterial\PycharmProjects\webstests\venv\lib\site-packages\websockets\server.py", line 726, in __await__
    server = yield from self._creating_server
TypeError: cannot 'yield from' a coroutine object in a non-coroutine generator
```
Quickly tested with a hacky try/except:
```py
    def __await__(self):
        try:
            server = yield from self._creating_server
        except TypeError:
            server = yield from self._creating_server.__await__()
```

PR #357, opened 2018-03-08
Ensure that code is not `None`

### LINKED ISSUE #350: TypeError("'<=' not supported between instances of 'int' and 'NoneType'",)
Hi!
I found a lot of errors in my log like:
```
2018-02-18 23:10:35,779 ERROR Task exception was never retrieved
future: <Task finished coro=<listen() done, defined at /opt/bet/src/server/websock/server.py:55> exception=TypeError("'<=' not supported between instances of 'int' and 'NoneType'",)>
  File "/opt/bet/src/server/websock/server.py", line 58, in listen
    msg = await websocket.recv()
  File "/opt/bet/env/lib/python3.6/site-packages/websockets/protocol.py", line 323, in recv
    raise ConnectionClosed(self.close_code, self.close_reason)
  File "/opt/bet/env/lib/python3.6/site-packages/websockets/exceptions.py", line 147, in init
    if 3000 <= code < 4000:
TypeError: '<=' not supported between instances of 'int' and 'NoneType'
```

My code:
```python
while True:
    msg = await websocket.recv()
    logger.debug("Got message from %s: %s", token, msg)
```

I use nginx for proxy websockets with this config file:
```
    location /websocket/ {
        proxy_pass http://127.0.0.1:8033;
        proxy_http_version 1.1;
        proxy_set_header Upgrade $http_upgrade;
        proxy_set_header Connection "upgrade";

        proxy_read_timeout 3600s;
        proxy_send_timeout 3600s;
    }
```

BeatButton:
Fixes #350

### DIFF
--- a/websockets/exceptions.py
+++ b/websockets/exceptions.py
@@ -141,6 +141,7 @@ class ConnectionClosed(InvalidState):
 
     """
     def __init__(self, code, reason):
+        code = code or 1006
         self.code = code
         self.reason = reason
         message = "WebSocket connection is closed: "


BeatButton:
Sorry about the force push shenanigans. I was in a rush and messed up.

aaugustin:
I'd like to understand why code is None when my logic doesn't expect it and fix that bug rather than just mask its symptoms.

BeatButton:
I don't see anything in WebSocketCommonProtoxol.recv that guarantees a close code being set, which is where the issue always surfaces in my reproductions. It doesn't seem like a terrible design choice, in my opinion, to assign "no reason" as the status code when no reason is given to the exception, but it's you're library, after all. Good luck bug stalking.

aaugustin:
Yes, your fix is pretty reasonable, I'd just like to make a quick effort to check there's not a larger issue lurking.

cjerdonek:
I think it might be better if the init method is allowed to set the code
attribute to None. That would do less to mask the underlying issue because
when diagnosing you’d be able to make a distinction between the exception
being raised with code None versus raising with code 1006. If there’s ever
a time where a number is really needed, you could change None to 1006 then.

Gorialis:
If it's any help, I experience this error a lot when the connection dies due to a node (such as a router or proxy) dropping out, severing the connection before the server could communicate a close code. I don't know if this covers all cases, or even is the direct cause, but it might help if trying to reproduce this problem.

Gorialis:
Bumping this with a reminder that [3.4 accounted for None close codes](https://github.com/aaugustin/websockets/blob/3.4/websockets/exceptions.py#L72) so it seems unreasonable for 4.0 to not also do so.

Gorialis:
@aaugustin Would like an update on this, given the [reproducible test case in #350](https://github.com/aaugustin/websockets/issues/350#issuecomment-375881747) and the [dupe issue also expressing this problem](https://github.com/aaugustin/websockets/issues/368) I'm not sure why this issue is still on hold - especially with the aforementioned [3.4 behavior](https://github.com/aaugustin/websockets/blob/3.4/websockets/exceptions.py#L72) and the fact that this issue is fatal to wrappers forcing libraries to [downgrade in order to avoid it](https://github.com/Rapptz/discord.py/commit/e04629999ba32142ba11c20191fca5eed60930f3).

aaugustin:
I believe this is the same issue as #327, which started looking at on yesterday.

This issue isn't really on hold. I had other hobbies than maintaining it for the last few months and so did the other contributors, so nothing happened.

Quoting the README:

> Professional support is available if you — or your company — are so inclined. [Get in touch](https://fractalideas.com/).

That's a way to expedite resolution of this issue while allowing me to feed my family.

Gorialis:
The request was not to intimidate or make demand of you - my interests are to help resolve this issue as soon as possible such that many people, myself included, can continue to use this library reassured that our services won't be taken offline by unexpected issues.

I bumped this PR because I was concerned that with the delay that it may have been forgotten or ignored - that much has been cleared up, so I appreciate your efforts to explain the situation.

After some review and testing, I agree with you that #327 achieves fixing the same issue, and more elegantly than this PR does, so I'll be moving any further concerns there.

BeatButton:
Other PR is better, yeah.

aaugustin:
Thanks to the 13 people who thought thumbing-down my comment was going to help :-)

Is it really that offensive to suggest work might be exchanged money?

Getting paid for one's work is fairly common.

aaugustin:
I never considered getting paid for pressing a button. I wasn't planning to merge a change without a test. Writing one will likely take me one or two hours, because testing race conditions is often tricky.

Thanks for the kind words, though. I hope you're putting websockets to good use in your projects :-)

aaugustin:
For the benefit of whoever stumbles upon this discussion and isn't familiar with the economics of open-source, paying for small actions or symbolic rewards is a common way to reward all the work that was freely given away. In that context, "paying for pressing the merge button" could make sense, even if it isn't what I suggested here.

aaugustin:
Fixed in 402059e4a46a764632eba8a669f5b012f173ee7b (which answers the question of "how much work is needed to fix this issue?")

BeatButton:
You removed one SLOC and added two. So the answer, it seems, is "a trivial amount"
In hopes of shielding you from further embarrassment, I'm going to ignore this thread.

gaul:
The abuse in this issue discourages me and I hope participants reconsider how they treat others, especially those giving freely of their time.

@aaugustin Thank you for fixing this and all your hard work on websockets!

Gorialis:
Good to see this issue get a clean fix. Thanks for your time, and apologies for the brusque nature of this thread.

aaugustin:
@Gorialis No problem. I may have reacted too briskly to your latest message, which was entirely reasonable.

aaugustin:
@BeatButton I'm pretty sure we'd have a better discussion IRL than over GitHub comments. I agree that it's best to leave it here.

PR #374, opened 2018-04-28
Add support for HTTP Basic Auth in the client side.

### LINKED ISSUE #373: Support basic auth in client
It does not seem possible to connect to a server using basic auth?

For example:
```
websockets.connect('wss://myuser:mypass@myserver.com')
```

aaugustin:
Fix #373.

### DIFF
--- a/websockets/uri.py
+++ b/websockets/uri.py
@@ -15,13 +15,17 @@ from .exceptions import InvalidURI
 __all__ = ['parse_uri', 'WebSocketURI']
 
 WebSocketURI = collections.namedtuple(
-    'WebSocketURI', ['secure', 'host', 'port', 'resource_name'])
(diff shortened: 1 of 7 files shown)


websockets/uri.py
```diff
@@ -52,4 +54,7 @@ def parse_uri(uri):
     resource_name = uri.path or '/'
     if uri.query:
         resource_name += '?' + uri.query
-    return WebSocketURI(secure, host, port, resource_name)
+    user_info = None
+    if uri.username or uri.password:
+        user_info = uri.username, uri.password
```
messense: I think it would be clearer to write this with parentheses?

```python
use_info = (uri.username, uri.password)
```

aaugustin: OK, I'll make that change.

PR #380, opened 2018-05-01
Recv on closing connection

aaugustin:


websockets/protocol.py
```diff
@@ -313,15 +315,15 @@ def recv(self):
...
-        # Now there's no need to yield from self.ensure_open(). Either a
-        # message was received or the connection was closed.
-
         if next_message in done:
             return next_message.result()
         else:
             next_message.cancel()
             if not self.legacy_recv:
-                raise ConnectionClosed(self.close_code, self.close_reason)
+                assert self.state in [State.CLOSING, State.CLOSED]
```
messense: I wonder whether use a tuple instead of a list here is considered better practice?

```python
assert self.state in (State.CLOSING, State.CLOSED)
```

messense: Well, that does make sense.

RemiCardona:
On our pre-prod servers, we hit this assertion 3 times in 4 hours after the upgrade. I haven't done any sort of analysis, I'll dive into it next week.

aaugustin:
Aaaargh. We aren't out of the woods yet. I'm starting to think websockets doesn't track enough state to behave properly in all cases.

RemiCardona:
Just a quick heads up, websockets 5.0.1 has been in production for a couple weeks now and we hit this assert between 200 and 400 times per hour. So I've patched up websockets to - hopefully - help debug this.

My debugging effort is in this branch https://github.com/RemiCardona/websockets/tree/try_to_debug_assert I think the "per-object logging" is a useful feature which could also be extended to include `this.side` on protocol objects, but I haven't done so to keep things simple.

If any of this sounds useful to other websocket users/devs, I'll gladly shape this up into a proper PR.

PR #385, opened 2018-05-05
Explain how to secure websockets with TLS.

aaugustin:


docs/intro.rst
```diff
@@ -47,10 +48,40 @@ coroutine returns.
...
+
+.. literalinclude:: ../example/secure_client.py
+    :emphasize-lines: 13-14
+
+This example builds a context because it uses a self-signed certificate.
+
+To connect to a secure WebSocket server with a valid certificate (i.e. signed
+by a CA that your Python installation trusts), simply pass ``ssl=True`` to
+``connect`` instead of building a context.
+
```
parmentelat: My understanding is that for adapting the example to the case where a valid certificate is used, we would still need an `ssl_context` for the server side, and only the client side needs to be altered, right ?

If that's not what you mean to say, then the wording is confusing; if it is what you mean, I suggest that you make this a little more explicit, like with

>(i.e. *blabla*), alter the client code and simply pass ``ssl=True`` to ``connect`` instead of building a context.

example/secure_client.py
```diff
@@ -0,0 +1,22 @@
+#!/usr/bin/env python
+
```
parmentelat: Personally I find that adding a one-line comment to repeat the purpose of the file
could help a bit; in this particular case something like

    # the client side that runs over wss / TLS

example/secure_server.py
```diff
@@ -0,0 +1,24 @@
+#!/usr/bin/env python
+
```
parmentelat: ditto

    # server side, runs over TLS, with a self-signed certificate

example/secure_client.py
```diff
@@ -0,0 +1,22 @@
...
+import asyncio
+import os.path
```
parmentelat: totally unrelated to the PR, but generally speaking since you require python-3.4 I suggest you use `pathlib` instead of old-school `os.path`

docs/intro.rst
aaugustin: You're correct. I'll clarify.

example/secure_client.py
aaugustin: I'm old-school and haven't adopted pathlib yet — mostly because casting `Path` objects back to `str` for APIs that don't accept a `Path` is tedious. 

In this case, it works, so I'm making the change.

aaugustin: While we're talking of fancy new features, I'm going to use f-strings as well.

parmentelat: well f-strings comes with 3.6 only, so this sounds like too much in the case of your library since, IIUC, it runs on 3.4 and up

aaugustin: That's fine - many examples use features that aren't available in Python 3.4; I'm providing separate examples for older Python versions.

parmentelat:
Oh I see, these are just examples for the doc :)  
So yes of course then, go for f-strings at any rate !

Thanks !

PR #387, opened 2018-05-05
Add an example of health check.

### LINKED ISSUE #354: healtch check example
Hello,

can you please explain to me
how to serve websocket connections on some URL and HTTP requests on other URL?

My URLs looks like this

```
/ws     ... websocket
/health ... HTTP
```

I found some relevant information at
https://github.com/aaugustin/websockets/issues/311.
and
https://github.com/aaugustin/websockets/blob/master/websockets/test_client_server.py#L146

However, I am still not sure how to use the modified `WebSocketServerProtocol`  e.g. the `HealthCheckServerProtocol` correctly. See the snippet below.
The problem is that I do not know where to **register** the `HealthCheckServerProtocol` in my code:

```
class HealthCheckServerProtocol(WebSocketServerProtocol):
    async def process_request(self, path, request_headers):
        if path == '/__health__/':
            body = b'status = green\n'
            return OK, [('Content-Length', str(len(body)))], body

async def handler(wsp, path):
    print("I do not know how to get the wsp argument as an instance of HealtchCheckServerProtocol")

server = loop.run_until_complete(websockets.serve(handler, '0.0.0.0', '8080'))
```

Thank you!

PS: Will you be interested in adding such example to the example dir?
It seems to me that implementing health-check is fairly often

aaugustin:
Fix #354.

### DIFF
--- a/docs/deployment.rst
+++ b/docs/deployment.rst
@@ -75,4 +75,7 @@ widely different operational characteristics of HTTP and WebSocket.
 
 ``websockets`` provide minimal support for responding to HTTP requests with
 the :meth:`~server.WebSocketServerProtocol.process_request()` hook. Typical
-use cases include health checks.
+use cases include health checks. Here's an example:
+
+.. literalinclude:: ../example/health_check_server.py
+    :emphasize-lines: 9-13,19-20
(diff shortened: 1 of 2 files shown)


ddehghan:
I would add this.

   elif path not in ['/', '/socket']:
            return BAD_REQUEST, [('Content-Length', 0)], ""

This will prevent bots visiting the root of the domain from causing exceptions. 

Also it would be good to name the socket endpoint with en explicit name like /socket. this way you can filter out the traffic generated by bots to the root and other random urls. These will cause exceptions and operation headaches.

PR #391, opened 2018-05-12
Improve failing the WebSocket connection.

aaugustin:
Mainly:

- Attach the reason why the connection was closed to ConnectionClosed exceptions
- Better mirror the specification in the implementation and improve compliance slightly in some edge cases


aaugustin:
Unsure why test coverage isn't 100% on CircleCI. It is 100% locally.

aaugustin:
After rebasing on top of #402, I could bring test coverage back to 100%.

PR #393, opened 2018-05-12
Improve handling of HTTP requests.

aaugustin:
When receiving HTTP requests to WS endpoints:
- Avoid polluting logs at the warning level — but keep logs at the debug level because it must be possible to debug failures
- Return a HTTP 426 status code to plain HTTP requests


RemiCardona:
I'm late to the review party but :tada: yay for logging severity downgrade ! And the rest looks just fine too.

PR #413, opened 2018-05-24
Add support for Python 3.7.

### LINKED ISSUE #405: Add support for Python 3.7
Add it to the testing matrix when it's supported on CircleCI... For now I'm removing that change from #404.

aaugustin:
Hopefully for real this time.

This is annoyingly complicated.

Fix #405.

### DIFF
--- a/websockets/py35/server.py
+++ b/websockets/py35/server.py
@@ -1,0 +1,22 @@
+async def __aenter__(self):
+    return await self
+
+
+async def __aexit__(self, exc_type, exc_value, traceback):
+    self.ws_server.close()
+    await self.ws_server.wait_closed()
+
+
+async def __await_impl__(self):
+    # Duplicated with __iter__ because Python 3.7 requires an async function
+    # (as explained in __await__ below) which Python 3.4 doesn't support.
+    server = await self._creating_server
+    self.ws_server.wrap(server)
+    return self.ws_server
+
+
+def __await__(self):
+    # __await__() must return a type that I don't know how to obtain except
+    # by calling __await__() on the return value of an async function.
+    # I'm not finding a better way to take advantage of PEP 492.
+    return __await_impl__(self).__await__()
--- a/docs/changelog.rst
+++ b/docs/changelog.rst
@@ -8,6 +8,8 @@ Changelog
 
 *In development*
 
+* Added compatibility with Python 3.7.
+
 5.0
 ...
 
@@ -136,6 +138,8 @@ Also:
 
 3.3
 ...
+
+* Ensured compatibility with Python 3.6.
 
 * Reduced noise in logs caused by connection resets.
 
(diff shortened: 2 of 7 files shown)


websockets/py35/server.py
```diff
@@ -0,0 +1,21 @@
...
+    # Duplicated with __iter__ because __await__ must call coroutines with
+    # await in Python 3.7 while __iter__ must call them with yield from to
+    # preserve compatibility with Python 3.4.
+    server = await self._creating_server
+    self.ws_server.wrap(server)
+    return self.ws_server
+
+
+def __await__(self):
+    # I'm not finding a better way to take advantage of PEP 492.
```
cjerdonek: If you put the `async` in front of `__await__`, can you remove the `__await_impl__`?

aaugustin: Nope. I tried :-)

I found this solution on StackOverflow, where someone dug it from a lib written by the author of PEP 492.

I find this very weird.

If you had time to ask on the asyncio mailing-list (where you seem to post occasionally) if this is the intended way to write an `__await__` method, that would be very useful.

As far as I understand (i.e. not much), `__await__` needs to return an type that is most easily obtained as the return value of calling `__await__` on an `async def` function.

aaugustin: That type reports itself as `<class 'coroutine_wrapper'>` and I'm not sure what it is.

aaugustin:
I confirmed manually that tests pass under Python 3.7.0b2 with these changes.

websockets/py35/server.py
```diff
@@ -0,0 +1,21 @@
...
+
+async def __aexit__(self, exc_type, exc_value, traceback):
+    self.ws_server.close()
+    await self.ws_server.wait_closed()
+
+
+async def __await_impl__(self):
+    # Duplicated with __iter__ because __await__ must call coroutines with
+    # await in Python 3.7 while __iter__ must call them with yield from to
+    # preserve compatibility with Python 3.4.
```
cjerdonek: One more thing: I'm not sure this comment (and others like it in this PR) explain the situation completely accurately. For example, it doesn't say anything about why 3.6 works but 3.7 doesn't. I think you want to be saying something about the return types, which changed from 3.6 to 3.7.

aaugustin: Good point.

Thankfully all this is going away when Python 3.4 is EOL in about a year.

RemiCardona:
And I confirm that my project's code works with 5.0.1 *and* this PR, both with `PYTHONASYNCIODEBUG=1` and without, on both 3.4.2 (debian's python) and 3.6.5. When 5.0 came out I was banging my head to get our project to work on 3.4.2.

So, in any case, this PR is :+1: by me.

aaugustin:
@RemiCardona Does this PR improve the situation on 3.4.2? I'm not aware of issues with that version of Python, but if there are and this PR solves them, I can release it in version 5.1. (That would be nice after I announced support for Python 3.7 in 5.0 and removed it in 5.0.1.)

websockets/py35/server.py
aaugustin: I simplified the comments.

Issue #256, opened 2017-08-27
Reconsider get_header / set_header


aaugustin:
`get_header` was an interesting experiment but it makes it hard to work with multi-value headers.

`set_header` is symmetrical, for consistency it must be reconsidered as well.

aaugustin:
There's at least one public use of these APIs: https://github.com/channelcat/sanic/blob/e1c90202682d76361c77adb7bb2176730038466a/sanic/websocket.py#L58-L103

aaugustin:
Fixed in #423.

PR #423, opened 2018-06-04
Refactor handling of HTTP headers.

### LINKED ISSUE #210: Consider a better structure for HTTP headers
Follow up on a suggestion by @cjerdonek here: https://github.com/aaugustin/websockets/pull/200#discussion_r126314330

Currently websockets relies `http.client.HTTPMessage` because:

- due to an implementation leak, the public API used to document the `email.message.EmailMessage` type; `http.client.HTTPMessage` is also available in the stdlib, more appropriate, and API-compatible (it's a thin subclass)
- I didn't want to reinvent yet another variant of a multi-value dict — but I'm willing to reconsider this now that I dropped the stdlib for parsing HTTP headers

aaugustin:
Fix #210, #256.

### DIFF
--- a/docs/changelog.rst
+++ b/docs/changelog.rst
@@ -10,9 +10,33 @@ Changelog
 
 .. warning::
 
-    **Version 6.0 changes public APIs in the** :mod:`~websockets.handshake`
-    **module. If you're calling these APIs, you must update your code. This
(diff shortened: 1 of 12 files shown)


aaugustin:
I think this is ready to merge; feedback appreciated :-)

PR #438, opened 2018-07-15
Support yield from connect/serve on Python 3.7.

### LINKED ISSUE #435: Using yield from in Python 3.7 causes TypeError
Library what I use is using `yield from` with `websockets.connect` in Python 3.7 
```
yield from websockets.connect(self.address)
```
I am getting an error: `TypeError: cannot 'yield from' a coroutine object in a non-coroutine generator`

As I can see, in Python 3.7 `loop.create_connection` is now a native coroutine, so you can not `yield from` it. Maybe we can mark `__iter__` with `@asyncio.coroutine` not only for Python < 3.5?

NB: I am using master version websockets @ `39b050fc8`

aaugustin:
Fix #435.

### DIFF
--- a/websockets/client.py
+++ b/websockets/client.py
@@ -384,6 +384,7 @@ class Connect:
         self._creating_connection = loop.create_connection(
             factory, host, port, **kwds)
 
+    @asyncio.coroutine
     def __iter__(self):                                     # pragma: no cover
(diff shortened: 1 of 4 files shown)


aaugustin:
I rebased the PR after updating to Circle CI 2.0 in order to be able to run tests on Python 3.7.

PR #449, opened 2018-07-29
Ping at regular intervals.

### LINKED ISSUE #383: Ping at scheduled intervals
We could add a `ping_interval` argument to the protocol.

When set, the protocol would start a task to send a ping at regular intervals.

This can help to keep connections open, especially across proxies with short timeouts.

This task would be a simple as:

```python
while self.open:
    yield from asyncio.sleep(self.ping_interval)
    yield from self.ping()
```

As a bonus, we could kill the connection if no pong is received in one `ping_interval`.

aaugustin:
Fix #383.


aaugustin:
This can't be merged yet because it's missing tests.

vgoklani:
Is it possible to control the text being passed in the ping: i.e. ping("keepalive")

aaugustin:
Thanks @bmoscon!

PR #482, opened 2018-09-29
Added a base websocket exception.

JakeMakesStuff:
This could be useful in situations where you want to catch all websocket exceptions.

### DIFF
--- a/src/websockets/exceptions.py
+++ b/src/websockets/exceptions.py
@@ -1,4 +1,5 @@
 __all__ = [
+    'WebsocketError',
     "AbortHandshake",
     "ConnectionClosed",
     "DuplicateParameter",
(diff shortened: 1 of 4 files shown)


JakeMakesStuff:
Hmm, not quite sure why Circle CI failed.

src/websockets/exceptions.py
```diff
@@ -20,7 +21,14 @@
 ]
 
 
-class InvalidHandshake(Exception):
+class WebsocketError(Exception):
```
messense: Should it be `WebSocketError`?

aaugustin: For consistency with other APIs, yes.

JakeMakesStuff:
Sorry for the merge conflict, I just pushed a fix for that.

aaugustin:
Thanks for your contribution!

I wanted to take this opportunity to rationalize the exception tree, which I just did in 8f1db07bb81e60c10e1d0a3efedfbbeb08962ac5.

You'll have `WebSocketException` in version 8.0, which I hope to release within a few days.

JakeMakesStuff:
Thanks @aaugustin, it's appreciated for the workflow which I have. :)

PR #488, opened 2018-10-17
Update README.rst

vkli:
add link to a tutorial for users without prior WebSockets experience

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -36,6 +36,10 @@ Python with a focus on correctness and simplicity.
 Built on top of ``asyncio``, Python's standard asynchronous I/O framework, it
 provides an elegant coroutine-based API.
 
+You will need a basic understanding of how to use WebSockets (the `Socket.IO Tutorial`_ is an excellent starting point).
+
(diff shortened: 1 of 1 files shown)


aaugustin:
I'm afraid this could cause confusion because Socket.IO does a lot more than websockets.

aaugustin:
See also: https://socket.io/docs/#What-Socket-IO-is-not

aaugustin:
Linking to a good websockets tutorial would be a good idea.

I would prefer something clearly focused on websockets - no more, no less - so there's no risk of confusion between websockets and everything else Socket.IO does.

I'll take a second look.

aaugustin:
I looked again and I'm still uncomfortable linking to this tutorial.

It demonstrates an event-driven approach which websockets doesn't use — it provides a higher-level abstraction. Several people filed issues because they were confused not to find an event-driven API in websockets. Making them read this tutorial could increase their confusion.

I've been looking for other tutorials and I'm not finding something I'm really wanting to link to :-(

Issue #470, opened 2018-09-08
Prevent running recv() concurrently


aaugustin:
I think it's documented as being possible but it's a Bad Idea because it leads to non-deterministic behavior. Per the Do Not Give Users Guns Aimed At Feet principle, it should raise `RuntimeError`, like `asyncio.StreamReader.read` and friends do.

Issue #459, opened 2018-08-19
test_protocol.py doesn't provide 100% test coverage of protocol.py


aaugustin:

aaugustin:
Since this is now legacy, I care less, and I think I'll remove that code before making this improvement.

PR #506, opened 2018-10-30
Rewrite recv() to avoid a data loss bug.

### LINKED ISSUE #486: asyncio.wait_for(ws.recv(), timeout=10) drop messages
Hi guys, i'm checking a strange behaviour

using "asyncio.wait_for(ws.recv(), timeout=10)" something drop messages, i don't know why, but anyone know something about it or have a better method to recv() without dropping messages?

maybe use recv() with a new parameter timeout to asyncio.wait() function?

aaugustin:
A race condition between receiving a message and cancelling recv() could
result in losing the message.

Fix #486 and #470.


aaugustin:
CI fails because there's no coverage for `max_queue`. This could be related to #459.

aaugustin:
@lgrahl Once this is done, I'll do one last review of cancellation handling and then I can release version 7.0. Sorry for the long wait.

PR #528, opened 2018-12-13
Handle redirects in client when connecting

johnsca:
Per https://tools.ietf.org/html/rfc6455.html#section-4.2.2 the server may redirect the client during the handshake. This allows the client to handle redirects properly instead of raising an InvalidStatusCode error.


aaugustin:
This is a good idea. Thanks for submitting this pull request.

I need to familiarize myself a bit more with the implementation before I merge it.

I may want to tweak the patch a bit e.g. I haven't used _-prefixed variables in this library until now.

aaugustin:
Upon further thought, if you add a line in the changelog mentioning this improvement, I'll merge the PR as is. It works and it's well tested. No need to wait :-)

PR #530, opened 2018-12-14
Enable GNU Readline for interactive client

lwzm:


aaugustin:
This seems like a reasonable change. However I'm failing to see a change in behavior with this patch. Perhaps my Python installations load `readline` automatically? Would you mind clarifying under which circumstances this is useful?

lwzm:
All of my development enviroments include:
python3.7 (anaconda) in Linux(CentOS 7, Linux 3.10);
python3.7 (anaconda) in Linux(CentOS 6, Linux 2.6.32);
python3.7 (anaconda) in macOS (10.13.6, Darwin 17.7.0);
python3.7 (brew install) in macOS (10.13.6, Darwin 17.7.0);

counld use this patch.

And i try to install from official package(python-3.7.1-macosx10.9.pkg), start
`python -m websocket ws://xxxxx`,
python does not import readline.

all above test cases, python do import readline automatically if only start a pure interactive shell via press `python<Enter>`.

check has readline already loaded:
>>> 'readline' in sys.modules

or

$ lsof -p  SOME_PY_PID

aaugustin:
OK, let's add this!

For consistency with the rest of the module, I would suggest:

1. to drop the `isatty` conditions — the module is only designed for interactive use; let's not suggest it could work in other circumstances (that would require more work)
2. to prevent flake8 from complaining with the unused import with `import readline  # noqa`
3. to move the code at the very beginning of the `main()` function, as this is quite similar in purpose to `win_enable_vt100()`.

If you'd like to update the pull request with these changes, I'll merge it.

(I can also make them myself if you're lacking time / motivation to do so.)

lwzm:
Follow your suggestion, and happy new year!

PR #544, opened 2019-01-02
Let user override host/port from what is in the uri when connecting

### LINKED ISSUE #540: Allow override for host/port parameters
What
-----

Say I want to connect to a TLS end-point, but can only do it through an a ssh tunnel

1. I'll start a tunnel like so:

```
ssh -N -L 50321:echo.websocket.org:443 ssh.jumper.host
```

2. Then connect `echo.websocket.org:443` via `localhost:50321`

```python
async with websockets.connect('wss://echo.websocket.org', 
                              host='127.0.0.1', 
                              port=50321) as websocket:
    await websocket.send('test')
    ...
```

Right now this code won't work, but I can achieve the same result by supplying connected socket and an extra `server_hostname` parameter. It is quite a bit more work and I'd rather let underlying `create_connection` code setup socket of its own.

Other use-case is supplying IP address of a server, a kind of in app DNS caching.

How
----

It doesn't look like this requires many changes. I have prototyped this here: https://github.com/Kirill888/websockets/commit/a9f8846c7958dfabc41d3e996d47735783f2b5d8

Happy to open a PR if you are happy with such change.

Kirill888:
In some cases it is useful to connect using an IP address of a server, but you
might still need to keep proper hostname in the url for TLS or for vhosts
routing. For example if you are connecting to `wss://example.com` but via an ssh
tunnel you might want to set `host='127.0.0.1', port=50032` but then supply
`server_hostname='example.com'` for TLS validation.

With this change connection logic detects that alternative `host|port` is
supplied by the user and does not attempt to override it with values from the
uri. If secure connection is requested and host override is used then also set
`server_hostname` to the host name in the uri (unless `server_hostname` was
already supplied by the user)

- closes #540

### DIFF
--- a/src/websockets/client.py
+++ b/src/websockets/client.py
@@ -328,6 +328,11 @@ class Connect:
     a ``wss://`` URI, if this argument isn't provided explicitly, it's set to
     ``True``, which means Python's default :class:`~ssl.SSLContext` is used.
 
+    Normally ``host`` and ``port`` parameters of
+    :meth:`~asyncio.BaseEventLoop.create_connection` are populated from the
+    supplied URI, but can be overwritten by the caller if desired, this can be
+    useful when connecting via ssh tunnel for example.
+
     The behavior of the ``ping_interval``, ``ping_timeout``, ``close_timeout``,
     ``max_size``, ``max_queue``, ``read_limit``, and ``write_limit`` optional
     arguments is described in the documentation of
@@ -461,18 +466,20 @@ class Connect:
             subprotocols=self._subprotocols,
             extra_headers=self._extra_headers,
         )
+        kwds = self._kwds.copy()
 
-        if self._kwds.get("sock") is None:
-            host, port = self._wsuri.host, self._wsuri.port
-        else:
-            # If sock is given, host and port mustn't be specified.
-            host, port = None, None
+        if kwds.get("sock") is None:
+            kwds.setdefault("host", self._wsuri.host)
+            kwds.setdefault("port", self._wsuri.port)
+
+        if kwds.get("ssl") and kwds.get("host") != self._wsuri.host:
+            kwds.setdefault("server_hostname", self._wsuri.host)
 
         self._wsuri = self._wsuri
         self._origin = self._origin
 
         # This is a coroutine object.
-        return self._loop.create_connection(factory, host, port, **self._kwds)
+        return self._loop.create_connection(factory, **kwds)
 
     @asyncio.coroutine
     def __iter__(self):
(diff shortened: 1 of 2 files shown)


aaugustin:
This PR touches the same area as #229. In that issue, the goal was to let the user provide an already-configured socket.
Back then we deciding against

aaugustin:
This PR touches the same area as #229. In that issue, the goal was to let the user provide an already-configured socket.

Back then, we deciding against making `server_hostname` default to `wsuri.host` like this PR does, but I'm open to reconsidering this choice in the name of convenience.

I would like to review how `asyncio` handles precedence of the various arguments accepted by `create_connection`, to make sure websockets adopts the same order.

Is there a specific reason why you made a copy of `self._kwds` instead of mutating it? Or was it just a matter of not modifying something you weren't sure you could modify?

Finally I've been thinking about the situation where the server redirects the connection. `websockets` attempts to follow the redirect — this is a new feature in the master branch, not yet released. It seems to me that this feature is incompatible with connection overrides (either with ``host + port`` or with ``sock``). If you need to control exactly where you're connecting, I assume you aren't expecting to be redirected somewhere else. This deserves documentation at a minimum, perhaps a warning or even an exception. It's more of a follow-up to the patch that added redirection so I'm going to file a separate issue for that.

Kirill888:
@aaugustin 

>Is there a specific reason why you made a copy of self._kwds instead of mutating it? Or was it just a matter of not modifying something you weren't sure you could modify?

I thought injecting `server_hostname` into `self._kwds` was probably not ok, but I probably shouldn't have moved `host` and `port` in there like I did.

Maybe it would be clearer if I didn't copy `self._kwds`, but instead created extra dictionary with options like `server_hostname`, and did explicit check for presence of those in the user supplied options instead of relying on the behaviour of `.setdefault` function.

I think it's reasonable to demand that user supply `server_hostname` explicitly rather than copying it from the uri, but also don't see any harm in doing it automatically, note you can still override it by explicitly supplying `server_hostname` . The most important part is that if `host` parameter is supplied by the user it should be used in preference to what's in the uri. I can not just put "the proxy ip" into uri and then use `server_hostname` for SSL validation as this will break access to servers behind vhosts proxy.

Not sure about redirects, I would say it's probably ok to let that break if user supplies connected socket or host:port, which is essentially the same thing, but with more convenient interface for simple cases like ssh forward proxy.

aaugustin:
Sounds good. I've been working on refactoring the redirect implementation, then I'll update this patch.

aaugustin:
Superseded by #632. I tried to start from your proposal and make minimal changes. Unfortunately I ended up with a complete rewrite. Sorry!

Kirill888:
yeah, no worries, thanks for doing this

Issue #541, opened 2019-01-01
Wait for connections to terminate when closing a server


aaugustin:
Currently `WebSocketServer._close` calls `fail_connection(1001)` on all established connections. This sends a close frame with code 1001 and stops processing data immediately.

It would be nicer to close connections properly with `close(1001)`and wait for a close frame from the remote endpoint.

This is a follow-up on #394.

ape364:
any news? maybe there is a workaround to solve this problem? i'm using `websockets==7.0`. 
if I understand correctly an exception must be raised in this case?

```
[2019-05-13 19:08:47,769] DEBUG    [websockets.protocol.write_frame:898] client > Frame(fin=True, opcode=9, data=b'\xd7\xb0\xd0\xf4', rsv1=False, rsv2=False, rsv3=False)
[2019-05-13 19:08:47,874] DEBUG    [websockets.protocol.read_frame:886] client < Frame(fin=True, opcode=8, data=b'\x03\xe9CloudFlare WebSocket proxy restarting', rsv1=False, rsv2=False, rsv3=False)
[2019-05-13 19:08:47,874] DEBUG    [websockets.protocol.write_close_frame:954] client - state = CLOSING
[2019-05-13 19:08:47,875] DEBUG    [websockets.protocol.write_frame:898] client > Frame(fin=True, opcode=8, data=b'\x03\xe9CloudFlare WebSocket proxy restarting', rsv1=False, rsv2=False, rsv3=False)
[2019-05-13 19:08:47,875] DEBUG    [websockets.protocol.eof_received:1220] client - event = eof_received()
[2019-05-13 19:08:47,876] DEBUG    [websockets.protocol.connection_lost:1229] client - event = connection_lost(None)
[2019-05-13 19:08:47,876] DEBUG    [websockets.protocol.connection_lost:1231] client - state = CLOSED
[2019-05-13 19:08:47,876] DEBUG    [websockets.protocol.connection_lost:1238] client x code = 1001, reason = CloudFlare WebSocket proxy restarting
[2019-05-13 19:08:47,876] DEBUG    [websockets.protocol.abort_keepalive_pings:1174] client - aborted pending ping: d7b0d0f4
[2019-05-13 19:08:47,877] DEBUG    [websockets.protocol.close_connection:1053] client x closing TCP connection
```

PR #552, opened 2019-01-11
Call the close() method to terminate websocket connections

lburg:
Simple change to fix ticket #541.

Is this all there is to it?


aaugustin:
Superseded by #641 because I wanted to do things a bit differently. Thanks for the patch!

PR #558, opened 2019-01-19
Add static typing with mypy.

### LINKED ISSUE #553: Type hinting
It would be neat to add type hints for [static type checking with mypy](https://mypy.readthedocs.io/en/latest/introduction.html).

aaugustin:
Fix #553.


lgrahl:
Great (and fast) progress so far!

While you're at it, what do you think of adding `WebSocketServer` to `__all__` so it is easily importable for type hint purposes in projects using websockets?

src/websockets/framing.py
```diff
@@ -46,8 +61,24 @@
 # Using a list optimizes `code in EXTERNAL_CLOSE_CODES`.
 EXTERNAL_CLOSE_CODES = [1000, 1001, 1002, 1003, 1007, 1008, 1009, 1010, 1011]
 
-FrameData = collections.namedtuple(
-    "FrameData", ["fin", "opcode", "data", "rsv1", "rsv2", "rsv3"]
+
+Data = Union[str, bytes]
```
lgrahl: One of the types that may need exporting.

src/websockets/client.py
```diff
@@ -364,26 +372,26 @@ class Connect:
...
+        legacy_recv: bool = False,
+        klass: Type[WebSocketClientProtocol] = WebSocketClientProtocol,
+        timeout: float = 10,
+        compression: Optional[str] = "deflate",
+        origin: Optional[str] = None,
+        extensions: Optional[List[ClientExtensionFactory]] = None,
+        subprotocols: Optional[List[str]] = None,
+        extra_headers: Optional[HeadersLike] = None,
+        **kwds: Any
     ):
```
lgrahl: Missing `-> None`. There's an option to make mypy complain about these. IIRC it was `--disallow-untyped-defs`.

src/websockets/client.py
```diff
@@ -364,26 +372,26 @@ class Connect:
...
+        ping_interval: float = 20,
+        ping_timeout: float = 20,
+        close_timeout: Optional[float] = None,
+        max_size: int = 2 ** 20,
+        max_queue: int = 2 ** 5,
+        read_limit: int = 2 ** 16,
+        write_limit: int = 2 ** 16,
+        loop: Optional[asyncio.AbstractEventLoop] = None,
+        legacy_recv: bool = False,
+        klass: Type[WebSocketClientProtocol] = WebSocketClientProtocol,
```
lgrahl: Suggestion to use a [type variable with an upper bound](https://mypy.readthedocs.io/en/latest/generics.html?highlight=typevar#type-variables-with-upper-bounds):

Somewhere above...

`WSCP = TypeVar('WSCP', bound=WebSocketClientProtocol)`

... then, in the signature...

`klass: Type[WSCP] = WebSocketClientProtocol`

... and return `WSCP` instead of `WebSocketClientProtocol` in methods below.

This allows to return the actual type of the instance being created instead of the upper bound and prevents the user from needing to `cast`.

src/websockets/protocol.py
```diff
@@ -1001,18 +1020,18 @@ def writer_is_closing(self):
         """
         try:
             # Wait for the data transfer phase to complete.
-            if self.transfer_data_task is not None:
+            if hasattr(self, "transfer_data_task"):
```
lgrahl: Is this the result of a typing issue or is the change unrelated to adding types?

If it is related to typing, have you considered `cast` in this case?

src/websockets/server.py
```diff
@@ -752,31 +794,33 @@ class Serve:
 
     def __init__(
```
lgrahl: Suggest to use `WSSP = TypeVar('WSSP', bound=WebSocketServerProtocol)` here as well.

aaugustin:
Thanks for the review!

> A lot of the input parameters could be less constrained (see comments inline). From reading the typing docs and the mypy docs, it seems good practice to make input parameters as generic as possible while output parameters can be specific types (unless one believes this may change in the future).

That makes sense.

Probably the best way to improve the PR is to review which types I imported from `typing`, checking which one have generic equivalents, and updating the patch accordingly.

I remember that in some cases, I put `List` rather than `Iterable` to be able to iterate it more than once. However, `Sequence` would be a better choice!

I'll look at each comment later.

> If you wanted to, you could add a couple of more NewTypes for primitives such as str. For example, subprotocol could be a NewType('Subprotocol', str) to make the type distinct. Another example would be NewType('Compression', str). Frame could be split into EncodedFrame and DecodedFrame, etc.

Absolutely, that would make types more readable.

I considered this for `Subprotocol`. I didn't do it because I wasn't sure how the `Subprotocol` <--> `str` conversions would work when encoding or decoding HTTP headers (and, honestly, because I was lazy).

There's also a few tuples that come up several times and might benefit from dedicated types.

> I think types available in public signatures should also be exported. To not clutter __all__ with custom types and aliases, I've moved types into a separate typing module in my project. This may be applicable here as well.

Indeed, they become part of the public API so they should be exported.

I thought about a dedicated module for type definitions for three reasons:
- I was facing import loops and wondered if it could help — then I decided it wouldn't actually help
- I wanted to minimize the impact of typing on the code —  then I changed my mind and decided it was best to colocate types and code
- It could make sense to mirror how the stdlib declares types in a dedicated module — then noticed that most types were actual classes, not just types; importing them all in a module didn't seem all that useful

I'm hoping to improve the PR during the week-end :-)

lgrahl:
Another thing: Merging this will require Python 3.5.3 from then on. Python 3.5.2's `typing` module is too buggy:

```python
import asyncio
from typing import *
Optional['asyncio.Task[None]']
```

Will result in:

```
  File "/usr/lib/python3.5/typing.py", line 649, in __getitem__
    return Union[arg, type(None)]
  File "/usr/lib/python3.5/typing.py", line 552, in __getitem__
    dict(self.__dict__), parameters, _root=True)
  File "/usr/lib/python3.5/typing.py", line 512, in __new__
    for t2 in all_params - {t1} if not isinstance(t2, TypeVar)):
  File "/usr/lib/python3.5/typing.py", line 512, in <genexpr>
    for t2 in all_params - {t1} if not isinstance(t2, TypeVar)):
  File "/usr/lib/python3.5/typing.py", line 190, in __subclasscheck__
    self._eval_type(globalns, localns)
  File "/usr/lib/python3.5/typing.py", line 177, in _eval_type
    eval(self.__forward_code__, globalns, localns),
  File "<string>", line 1, in <module>
TypeError: 'type' object is not subscriptable
```

aaugustin:
Honestly I just want to drop Python 3.5 and stop thinking about all the things that only start working in Python 3.5.1 or 3.5.3 or who knows which 3.5.x...

aaugustin:
I rebased the PR on top of master after dropping Python 3.6.

src/websockets/client.py
```diff
@@ -49,19 +53,19 @@ class WebSocketClientProtocol(WebSocketCommonProtocol):
...
         *,
-        origin=None,
-        extensions=None,
-        subprotocols=None,
-        extra_headers=None,
-        **kwds
-    ):
+        origin: Optional[str] = None,
+        extensions: Optional[List[ClientExtensionFactory]] = None,
+        subprotocols: Optional[List[str]] = None,
```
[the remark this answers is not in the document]
aaugustin: Done, as well as every other List in an argument type. I kept them in result types.

src/websockets/client.py
```diff
@@ -523,17 +546,17 @@ def __await__(self):
...
+    connect = Connect
+
+else:  # pragma: no cover
 
     del Connect.__aenter__
     del Connect.__aexit__
     del Connect.__await__
 
-    async def connect(*args, **kwds):
+    async def connect(*args, **kwds):  # type: ignore
```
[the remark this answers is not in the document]
aaugustin: If memory serves, that didn't suffice to make mypy happy. Anyway that code is gone :-)

src/websockets/framing.py
aaugustin: I moved it to `websockets.typing` and documented it.

src/websockets/client.py
aaugustin: `--disallow-untyped-defs` is implied by `--strict`.

However declaring `-> None` is optional on `__init__` because that's always the expected return type.

aaugustin: I realize I included `-> None` for most `__init__` methods. I'm adding the missing ones.

src/websockets/client.py
aaugustin: I tried that; it causes errors I don't understand well and requests additional type annotations:

```
src/websockets/server.py:805: error: Incompatible default for argument "klass" (default has type "Type[WebSocketServerProtocol]", argument has type "Type[WebSocketServerProtocolType]")
src/websockets/server.py:865: error: Argument 1 to "WebSocketServerProtocol" has incompatible type "Callable[[WebSocketServerProtocolType, str], Awaitable[Any]]"; expected "Callable[[WebSocketServerProtocol, str], Awaitable[Any]]"
src/websockets/client.py:370: error: Incompatible default for argument "klass" (default has type "Type[WebSocketClientProtocol]", argument has type "Type[WebSocketClientProtocolType]")
src/websockets/client.py:509: error: Need type annotation for 'protocol'
src/websockets/client.py:534: warning: Returning Any from function declared to return "WebSocketClientProtocolType"
src/websockets/__main__.py:98: error: Need type annotation for 'websocket'
```

Not sure how to proceed from there.

src/websockets/protocol.py
aaugustin: Typing encouraged me to stop initializing attributes to `None` to avoid a plague of `Optional[...]` in type declarations.

It doesn't make a big difference in practice: if you use the attribute before it gets a valid value, you get "xxx has no attribute yyy" instead of "None has no attribute zzz" (when `xxx.yyy` = None).

src/websockets/server.py
aaugustin: See above — that didn't work well.

aaugustin:
For reasons unclear, type annotations appear inconsistenly in sphinx docs.

PR #568, opened 2019-02-03
Add ConnectionClosed subclass for normal closure.

### LINKED ISSUE #285: ConnectionClosed subclass for normal closure
I was thinking it would be a nice touch to add (and use) a `ConnectionClosed` subclass that corresponds to a normal closure (i.e. status code 1000).

This would make it easier for users to handle expected (normal) closures differently and more elegantly, namely without having to inspect the code attribute. For example, here is one place in the websockets code base that such a pattern could be used:

    try:
        while True:
            yield await self.recv()
    except ConnectionClosed as exc:
        if exc.code == 1000:
            return
        else:
            raise

Here is the code link:
https://github.com/aaugustin/websockets/blob/27549c4b390443b7504e937d4d974bd0855b4c7f/websockets/py36/protocol.py#L4

aaugustin:
Thanks @cjerdonek for the suggestion.

Fix #285.

### DIFF
--- a/src/websockets/protocol.py
+++ b/src/websockets/protocol.py
@@ -30,6 +30,8 @@ from typing import (
 
 from .exceptions import (
     ConnectionClosed,
+    ConnectionClosedError,
+    ConnectionClosedOK,
     InvalidState,
     PayloadTooBig,
     WebSocketProtocolError,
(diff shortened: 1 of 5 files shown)


src/websockets/protocol.py
```diff
@@ -653,6 +653,17 @@ def closed(self) -> bool:
 
     # Private methods - no guarantees.
 
+    @property
+    def connection_closed_exc(self) -> ConnectionClosed:
```
cjerdonek: Since this is creating a new exception object, I would make this a method rather than a property.

aaugustin:
I updated the PR, it should be ready to merge now.

aaugustin:
(GitHub shows a spurious commit because I forgot to push master before pushing the branch for this PR.)

PR #590, opened 2019-04-02
Mark package as typed

tyehle:
Just include the `py.typed` file so mypy knows it can use the annotations in the package source when typechecking other code. See [these docs](https://mypy.readthedocs.io/en/latest/installed_packages.html#making-pep-561-compatible-packages) for details. I'm not sure the change to zip_safe is really needed as I think it only applies to .egg packages, but the docs advised the change, so :man_shrugging:

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -51,7 +51,7 @@ setuptools.setup(
     packages=packages,
     ext_modules=ext_modules,
     include_package_data=True,
-    zip_safe=True,
+    zip_safe=False,
     python_requires='>=3.6',
     test_loader='unittest:TestLoader',
 )


aaugustin:
Thanks, I'll double check the docs quickly and merge this before the next version.

tyehle:
@aaugustin got any updates on this?

aaugustin:
I haven't been able to work on my open source projects for a couple months. I hope I'll have more time in May.

tyehle:
Hmmm, I just tried both and found the `py.typed` file in `/websockets-7.0/src/websockets/` for sdist and in `/websockets/` for bdist. I only tried running the build on my machine using python 3.7.3 though.

If you want a persistent test for this I'm not sure how to go about it. There might be a way to build a package and then unzip it and check for the file.

tyehle:
Oh woah, I just tried the build on OSX and the files aren't included :open_mouth: Looking into it.

tyehle:
Ok looks like I had some strange build state going on. The additions to those two files should do it though. See [this SO question](https://stackoverflow.com/q/7522250) for details about MANIFEST

tyehle:
Thanks! 🎉

PR #600, opened 2019-04-16
my changes

Morocat:
Allows multiple connections to use the same websocket. Only closes the socket once the last connection has dropped.

### DIFF
--- a/src/websockets/exceptions.py
+++ b/src/websockets/exceptions.py
@@ -262,3 +262,9 @@ class WebSocketProtocolError(Exception):
     Internal exception raised when the remote side breaks the protocol.
 
     """
+
+
+class TooManyOpenSocketsError(Exception):
+    """
+    Exception raised when too many clients try to connect.
+    """
--- a/src/websockets/server.py
+++ b/src/websockets/server.py
@@ -34,6 +34,7 @@ from .exceptions import (
     InvalidOrigin,
     InvalidUpgrade,
     NegotiationError,
+    TooManyOpenSocketsError,
 )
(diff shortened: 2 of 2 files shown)


aaugustin:
I'm seeing two things in this PR:

- You added the ability to limit the number of connections. I can see how this could be useful. This feature should have its own PR with docs and tests. Note that HTTP 429 isn't an appropriate status code because you don't know if all connections come from the same user.

- You added a bunch of `if len(self.ws_server.websockets) == 1` conditions. I'm not sure what you're trying to achieve. The title of the PR ("Multiple connections support") suggests that you believe that websockets currently doesn't support multiple connection. It does; I already opened more than 10k connections to a single process.

Morocat:
The `if len(self.ws_server.websockets) == 1` checks are designed around the fact that when I closed one connection the rest of them also closed. These prevent the websocket from actually closing until the last connection has closed.

What would an appropriate status code for too many connections be?

aaugustin:
I'm afraid you're confused: with the current implementation, the server remains alive until you close it, regardless of how many connections open or close.

Morocat:
In order to test this I created a javascript client where I opened a websocket to my python app target. From the same browser I opened two tabs with the javascript webpage and streamed data from the python app to the webpage. Once I closed one of the tabs, the other tab lost its websocket connection as well. The python websocket server continued to run but the websocket that was hosting both webpages had closed even though one of the web pages was still open.

aaugustin:
I don't understand "the websocket that was hosting both webpages". I ran similar tests and never faced such issues. I suspect an exception is raised somewhere in your code, causing the other connection to fail.

PR #636, opened 2019-07-01
Set reason in the close response for error 1011 when timing out waiti…

bsergean:
…ng for pong, and when there is an error in data transfer

Our client is disconnecting with error code 1011, and setting a reason when this happens helps troubleshooting the problem.

### DIFF
--- a/src/websockets/protocol.py
+++ b/src/websockets/protocol.py
@@ -794,10 +794,11 @@ class WebSocketCommonProtocol(asyncio.StreamReaderProtocol):
             # This shouldn't happen often because exceptions expected under
             # regular circumstances are handled above. If it does, consider
             # catching and handling more exceptions.
-            logger.error("Error in data transfer", exc_info=True)
+            reason = "Error in data transfer"
+            logger.error(reason, exc_info=True)
 
             self.transfer_data_exc = exc
-            self.fail_connection(1011)
+            self.fail_connection(1011, reason)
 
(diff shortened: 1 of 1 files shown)


bsergean:
Now my log looks like this, and I know that my disconnects are about ping timeouts. (I had read the documentation on that during the 7.0 release, but I thought that my library was handling ping/pong  correctly, but apparently isn't).

```
[2019-07-01 16:33:05.043] [info] Connection closed: Close code 1011 reason timed out waiting for pong
```

bsergean:
This breaks the unittest ... I'll try to adjust the tests to handle this extra reason, unless you think it isn't a good idea to send a reason here (I think it's a good thing ... to make client/server debugging easier).

aaugustin:
I think a larger scale effort is needed to clarify error messages. Overall, error messages in websockets are too low-level and don't help users.

In several recent issues, users aren't quite sure what's happening when they see a 1006 error, because no one knows it means the connection died.

Recently I improved exceptions in the opening handshake, the best example being https://github.com/aaugustin/websockets/commit/aa2a2bb52621626c5661f8be5de4985e18e87acf#diff-fa79765527e4efb131a2757cb3f3b1a1R103 when the TCP connection dies.

I should do something similar to clarify the reason why the connection fails.

bsergean:
Great. I don't think that this PR is very helpful in that current form for a 'full' so I'm tempted to close it to make some room.

If I can help in any way let me know.

PR #655, opened 2019-07-25
Update README.rst

jucie15:
modify require python version >= 3.6 to >= 3.6.1
cause 3.6.0 does not include Deque in typing. it was added on 3.6.1

Reference: https://docs.python.org/3/library/typing.html


balloob:
I help maintain @home-assistant which is a consumer of `websockets` hence I have an interest in this PR.

I would prefer if instead of updating the min version, `websockets` drops the Deque type and instead type it with `deque` imported from `collections` and keeps supporting Python 3.6.0.

aaugustin:
Thanks for the PR and sorry for not replying earlier!

Some significant code changes were required in addition to these documentation changes. I just made all these changes, which supersede this pull request.

PR #658, opened 2019-07-26
Update shutdown.py

hkaraoguz:
Added signal handler for "CTRL+C"

### DIFF
--- a/example/shutdown.py
+++ b/example/shutdown.py
@@ -18,5 +18,8 @@ loop = asyncio.get_event_loop()
 stop = loop.create_future()
 loop.add_signal_handler(signal.SIGTERM, stop.set_result, None)
 
+# For CTRL+C signal
+loop.add_signal_handler(signal.SIGINT, stop.set_result, None)
+
 # Run the server until the stop condition is met.
 loop.run_until_complete(echo_server(stop))
--- a/example/shutdown.py
+++ b/example/shutdown.py
@@ -14,12 +14,13 @@ async def echo_server(stop):
 
 loop = asyncio.get_event_loop()
 
-# The stop condition is set when receiving SIGTERM.
-stop = loop.create_future()
-loop.add_signal_handler(signal.SIGTERM, stop.set_result, None)
+try:
+    # The stop condition is set when receiving SIGTERM.
+    stop = loop.create_future()
+    loop.add_signal_handler(signal.SIGTERM, stop.set_result, None)
+    
+    # Run the server until the stop condition is met.
+    loop.run_until_complete(echo_server(stop))
 
-# For CTRL+C signal
-loop.add_signal_handler(signal.SIGINT, stop.set_result, None)
-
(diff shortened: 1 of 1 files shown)


aaugustin:
Catching `KeyboardInterrupt` would be more idiomatic to handle Ctrl-C, wouldn't it?

hkaraoguz:
Yes, the code is updated accordingly.

aaugustin:
I don't think this change is an improvement for three reasons:

1. The example no longer matches the text of the documentation where it's included (https://github.com/aaugustin/websockets/blob/34aaf6bcbbac62d8c605d5ba768709346ef87c6e/docs/deployment.rst#graceful-shutdown) because it doesn't close the server cleanly on SIGINT
2. It doesn't have the correct lines highlighted.
3. It demonstrates two things at once, which is a bad practice in documentation.

In fact I'm confused because I don't know what your goal is.

If you'd like to extend the section of the docs that talks axbout proper shutdown to discuss SIGINT handling, then you need to get into details about when Python does or doesn't register a default signal handler for SIGINT (it's a bit of a mess, as demonstrated by questions on StackOverflow), or point to another document that explains it (unfortunately the official Python docs fall a bit short).

I just rediscovered all this and remember that I sidestepped the problem by demonstrating SIGINT rather than SIGTERM. I'm not really willing to maintain a discussion of SIGINT vs. KeyboardInterrupt in Python.

hkaraoguz:
The goal of this change is to be able to exit gracefully when the script receives a KeyboardInterrupt (CTRL+C). 
The current example script does not shutdown gracefully when it gets KeyboardInterrupt. At least it did not work for me (Python 3.7 in Ubuntu 18.04).
1. In order to fix this, I first used the example code as is and added a line to capture SIGINT. It worked for me and I created the pull request. 
2. Following your advise, I switched to capturing KeyboardInterrupt. A similar solution is given here https://www.roguelynn.com/words/asyncio-graceful-shutdowns/. This solution also worked for me.

Of course it is up to you to merge this change but as I said before, in it's current state, the script is not shutting down gracefully when it receives CTRL+C. If you decide to merge, I can update the documentation accordingly.

aaugustin:
To clarify my point:

- Overriding the default signal handler for SIGINT (which was your initial proposal) makes it easy to create a Python program that runs in a loop and that you cannot exit with Ctrl-C — that's why Python translates SIGINT to KeyboardInterrupt in the first place.

- Triggering a KeyboardInterrupt exception to exit is prone to race conditions; catching it is even worse; it's best to rely on a signal handler that sets a flag that you can check later.

See https://vorpus.org/blog/control-c-handling-in-python-and-trio/#the-precedent-control-c-in-regular-python for a longer discussion of this and https://github.com/python-trio/trio/blob/master/trio/_core/_ki.py for a sense of where we'd end if we went to the bottom of this discussion.

In the end, I stand by my choice to demonstrate only SIGTERM. I'll clarify the docs slightly.

PR #662, opened 2019-08-12
fix: permit None in type annotations

gthb:
Fix type annotations for four parameters which are documented to accept `None`.


aaugustin:
Good catch, thanks!

PR #664, opened 2019-08-26
Support Python 3.6.0

balloob:
The `Deque` type from `typing` was added to Python 3.5.4 and Python 3.6.1. It is not available in Python 3.6.0 ([Source](https://docs.python.org/3/library/typing.html#typing.Deque)). By relying on `Deque`, the minimum version of `websockets` has become 3.6.1 while the readme mentions 3.6.0. If this was by choice, close this PR.

Closes #655

### DIFF
--- a/src/websockets/protocol.py
+++ b/src/websockets/protocol.py
@@ -261,7 +261,7 @@ class WebSocketCommonProtocol(asyncio.StreamReaderProtocol):
         self.connection_lost_waiter: asyncio.Future[None] = loop.create_future()
 
         # Queue of received messages.
-        self.messages = collections.deque()
+        self.messages: collections.deque = collections.deque()
         self._pop_message_waiter: Optional[asyncio.Future[None]] = None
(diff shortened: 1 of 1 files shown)


balloob:
MyPy is not happy

```
src/websockets/protocol.py:264: error: Implicit generic "Any". Use "typing.Deque" and specify generic parameters
src/websockets/protocol.py:493: warning: Returning Any from function declared to return "Union[str, bytes]"
```

A solution could be to add `# type: ignore`

aaugustin:
I'm starting to think it would be best to document that at least 3.6.1 is required:

- generally speaking, I think it's best to stick to the latest 3.6.x micro version (or 3.7.y, etc.);
- that's what hosted CI services do; I don't have the resources or the will to run CI on every micro version;
- there's a few other places in the code base where 3.6.0 holds us back (look for "3.6.1" in comments).

aaugustin:
If you don't mind, I'll keep it open so I don't forget to update the documentation, then I'll close it.

PR #680, opened 2019-10-08
Hardcode top-level exports.

### LINKED ISSUE #679: mypy: Module has no attribute "connect"  [attr-defined]
As of mypy 0.730 using `no_implicit_reexport = True` errors are reported when importing names from the top-level module. I'm guessing this is due to mypy not being able infer the dynamic contents of `__all__`, and instead having relied on inferring the imports previous to 0.730.

Would you be open to a pull request that hard-codes the contents of `__all__` instead of building it dynamically from the __all__s of submodules?

antonagestam:
fixes #679 

Also adds a test to make sure no two names are competing with each other in the top-level namespace (e.g. two sub-modules can't export the same name).

### DIFF
--- a/src/websockets/__init__.py
+++ b/src/websockets/__init__.py
@@ -1,6 +1,5 @@
 # This relies on each of the submodules having an __all__ variable.
 
-from . import auth, client, exceptions, protocol, server, typing, uri
 from .auth import *  # noqa
 from .client import *  # noqa
 from .exceptions import *  # noqa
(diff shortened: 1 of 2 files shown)


src/websockets/__init__.py
```diff
@@ -1,6 +1,6 @@
 # This relies on each of the submodules having an __all__ variable.
 
-from . import auth, client, exceptions, protocol, server, typing, uri
+from . import auth, client, exceptions, protocol, server, typing, uri  # noqa
```
aaugustin: I don't think this line serves any purpose anymore. Could you remove it?

tests/test_exports.py
```diff
@@ -0,0 +1,22 @@
...
+    + websockets.uri.__all__
+)
+
+
+class TestExportsAllSubmodules(unittest.TestCase):
+    def test_top_level_module_reexports_all_submodule_exports(self):
+        self.assertEqual(set(combined_exports), set(websockets.__all__))
+
+    def test_submodule_exports_are_globally_unique(self):
+        self.assertEqual(len(set(combined_exports)), len(combined_exports))
```
aaugustin: Good idea.

antonagestam: I've been burned by this once before, test cases were not being run because they where clashing in a namespace 😐

PR #684, opened 2019-10-31
Fix up type hints on optional arguments

alexcoplan:
The arguments ping_interval, ping_timeout, max_size, and max_queue can all be None, so the type hints should be Optional[T].

The second commit here fixes a type error regarding the basic auth username/password.


alexcoplan:
Looks like it's now just formatting changes that are upsetting the CI. Would you like the automatically-reformatted code in a third commit?

aaugustin:
For the styling changes, run `make style`, fix any reported problems, and commit the result.

`make style` formats the code automatically with `black`.

aaugustin:
The first commit looks fine.

The second commit doesn't make sense to me. If `username` or `password` is `None`, then `build_authorization_basic` shouldn't be called at all; it shouldn't accept `None` arguments. Let's keep the representation of "no basic auth info" as `wsuri.user_info is None`, rather than `wsuri.user_info == (None, None)`.

aaugustin:
Type errors around basic auth credentials appeared with mypy 0.740. mypy 0.730 didn't detect them. This problem is now fixed in the master branch.

alexcoplan:
I'll close this PR and make a new one with just the type hint fixes.

PR #685, opened 2019-11-04
Fix type hints on optional client/server args

alexcoplan:
Make ping_interval et al. optional so that code that passes None here will type check. Supersedes #684


aaugustin:
Thanks!

Issue #683, opened 2019-10-30
cibuildwheel is not running the test suite


jlaine:
I noticed that the cibuildwheel run for Linux + Windows is not exercising the test suite:

Linux:

```
+ sh -c 'python3 -W default -m unittest'

~ /project

----------------------------------------------------------------------

Ran 0 tests in 0.000s
```

Windows:
```
+ python -W default -m unittest
Using default MSVC build environment for 64 bit architecture
Executing: python -W default -m unittest
----------------------------------------------------------------------
Ran 0 tests in 0.000s
OK
```

jlaine:
After some poking around I believe the `.appveyor.yml` needs to be amended with:

```
 CIBW_TEST_COMMAND: python -W default -m unittest discover -s {project}/tests
```

aaugustin:
This might be an accident dating back to when I split src/ and tests/.

Or it might be something I decided in the past, because I don't care much about running tests with every built wheel.

No strong opinions there, happy to merge a PR if someone has an opinion :-)

jlaine:
I'm now almost certain it is a result of splitting `src/` and `tests/`, I encountered the exact same thing for another package. I'll put together a PR.

aaugustin:
I've been reconsidering this today as I moved the CI to GitHub actions. I don't think it's really necessary to run tests for each wheel, so I'm not going to try.

PR #686, opened 2019-11-04
Ensure test suite is run by cibuildwheel (fixes: #683)

jlaine:


jlaine:
Hm, well that failed :(

aaugustin:
Perhaps `python -W default -m unittest discover -s {project}` (without `/tests`) would work. Then relative imports would be possible inside the `tests` package.

aaugustin:
I'm taking a shot at this in #806.

Issue #653, opened 2019-07-23
Documentation: How can server broadcast to all connected clients at regular interval?


felix-pb:
I'm trying to implement a WebSocket server that, once it's ready to accept connections, starts generating data periodically. Whenever data is generated (e.g. every 2 seconds), I want to broadcast the data to all connected clients.

I looked at the documentation and could not find an example like that. There is an example where the WebSocket server sends a message periodically to connected clients, but each client has its own "interval" (i.e. all clients won't receive the message at the same time).

It would be great if the documentation had such an example. I'm coming from Node.js and this is pretty easy to do with the 'ws' npm package. I'm new to Python, so I hope my question is not ridiculously stupid.

aaugustin:
You need to:

- keep track of connected clients e.g. like this https://websockets.readthedocs.io/en/stable/intro.html#registration
- start an asyncio task that broadcasts the data; this task must in parallel to your server

lgrahl:
Well, it's certainly not a stupid question but I'd say it's more of an asyncio question, so I'm not sure it needs to be added to the docs. But here's how it could be done:

```python
import asyncio

class MessageGenerator:
    def __init__(self):
        self.message = asyncio.Future()

    async def run(self):
        while True:
            await asyncio.sleep(1.0)
            self.message.set_result(b'meow')
            self.message = asyncio.Future()


async def handler(name, message_generator):
    while True:
        message = await asyncio.shield(message_generator.message)
        print(name, message)


async def main():
    message_generator = MessageGenerator()
    asyncio.create_task(message_generator.run())
    asyncio.create_task(handler('1', message_generator))
    asyncio.create_task(handler('2', message_generator))
    await asyncio.sleep(10.0)

asyncio.run(main())
```

Your server can now just do what the `handler` function does in its client handler function.

aaugustin:
@lgrahl I'm not sure I understand your example.

Here's the sort of thing I had in mind (completely untested):

```python
CLIENTS = set()

async def broadcast():
    while True:
        for ws in CLIENTS:
            await ws.send("woof")
        await asyncio.sleep(2)

asyncio.create_task(broadcast())

async def handler(websocket, path):
    CLIENTS.add(websocket)
    try:
        async for msg in ws:
            pass
    finally:
        # Unregister.
        CLIENTS.remove(websocket)

start_server = websockets.serve(handler, ...)

asyncio.get_event_loop().run_until_complete(start_server)
asyncio.get_event_loop().run_forever()
```

[...]

PR #689, opened 2019-11-11
Add a broadcast client(s)/server example.

HazenBabcock:
This pull request adds a broadcast server (and clients) example as well as the documentation as described in issue #653 .

### DIFF
--- a/example/broadcast_server.py
+++ b/example/broadcast_server.py
@@ -1,0 +1,36 @@
+#!/usr/bin/env python
+import asyncio
+import time
+import websockets
+
+async def broadcast():
+    while True:
+        message = str(time.time())
+        await asyncio.gather(
+            *[ws.send(message) for ws in CLIENTS],
+            return_exceptions=False,
+        )
+        await asyncio.sleep(2)
+
+async def handler(websocket, path):
+    CLIENTS.add(websocket)
+    try:
+        async for msg in websocket:
+            pass
+    except websockets.ConnectionClosedError:
+        print("Connection closed")   
+    finally:
+        CLIENTS.remove(websocket)
+
+CLIENTS = set()
+
+loop = asyncio.get_event_loop()
+loop.create_task(broadcast())
+
+start_server = websockets.serve(handler, "localhost", 8765)
+
+loop.run_until_complete(start_server)
+try:
+    loop.run_forever()
+except KeyboardInterrupt:
+    print("Exiting")
--- a/docs/intro.rst
+++ b/docs/intro.rst
@@ -109,6 +109,22 @@ Then open this HTML file in several browsers.
 .. literalinclude:: ../example/counter.html
    :language: html
 
+Broadcast example
+-----------------
+
+Here's a WebSocket broadcast server example.
+
+The server keeps track of the current clients. It broadcasts the current time
+every two seconds.
+
+.. literalinclude:: ../example/broadcast_server.py
+
+Here's the corresponding WebSocket client example.
+
+.. literalinclude:: ../example/broadcast_client.py
+
+It listens until the user exits with crtl^c.
+
 Common patterns
 ---------------
 
--- a/docs/intro.rst
+++ b/docs/intro.rst
@@ -123,7 +123,7 @@ Here's the corresponding WebSocket client example.
 
 .. literalinclude:: ../example/broadcast_client.py
 
-It listens until the user exits with crtl^c.
+It listens until the user exits with ctrl^c.
 
 Common patterns
 ---------------
--- a/example/broadcast_client.py
+++ b/example/broadcast_client.py
@@ -1,0 +1,18 @@
+#!/usr/bin/env python
+import asyncio
+import websockets
+
+async def client():
+    uri = "ws://localhost:8765"
+    async with websockets.connect(uri) as websocket:
+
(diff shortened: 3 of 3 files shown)


example/broadcast_server.py
```diff
@@ -0,0 +1,36 @@
...
+import asyncio
+import time
+import websockets
+
+async def broadcast():
+    while True:
+        message = str(time.time())
+        await asyncio.gather(
+            *[ws.send(message) for ws in CLIENTS],
+            return_exceptions=False,
```
hilbix: Haven't tested this, but IMHO there should be `return_exceptions=True`, as from the docs: "exceptions are ... aggregated in the result list".  (We probably do not want to see `ConnectionClosed` exceptions propagated into `broadcast()`.)

example/broadcast_server.py
```diff
@@ -0,0 +1,36 @@
+#!/usr/bin/env python
+import asyncio
+import time
+import websockets
+
+async def broadcast():
+    while True:
+        message = str(time.time())
+        await asyncio.gather(
```
hilbix: `await`?  This makes `broadcast()` sleep until `asyncio.gather()` is fullfilled!  Haven't testet this, but IMHO if one client hangs, so `ws.send()` takes ages to finish, `broadcast()` will sit there and wait.  Perhaps forever.  This might be wanted .. or rather not.  I am still learning AsyncIO but this exactly is the most common pattern which most commonly annoys me most:  How to properly express a fire-and-forget-pattern in AsyncIO which is expressed in bash as trivial as `( something & )`?!?  It must be done safely, as else GC might finalize the objects or we have some memory leak!  The entry point probably is `task = asyncio.ensure_future(asyncio.gather(...` (to be Python 3.6 compatible).  But this is only a third of the solution.  It now leaves us (only me?) with a dangling `task` object which now must be stuffed somewhere safe to keep it away from GC. And when the task is finished, this stuffing must be taken back to free memory again, perhaps with some PITA like `task.add_done_callback()`.  (Didn't I just not come across the correct pattern in AsyncIO for this yet, or is it really so hard to express in AsyncIO?)

example/broadcast_server.py
HazenBabcock: Not sure, I think it is this way so that exceptions immediately propagate into `handler` which removes the corresponding websocket from the `CLIENTS` set. However it has been a while and my understanding of `asnycio` is poor. My testing (as I recall) consisted of running several clients and verifying that if I killed one with CTRL-C everything else kept running as expected.

example/broadcast_server.py
hilbix: OK, forget this.  Just remove the `await` and it should work, as according to my tests (Ubuntu 18.04) one can call `asyncio.gather()` to actually run the awaitables.

Also the tests shows, that `asyncio.ensure_future(g)` just returns `g` when `g = asyncio.gather()`.  Hence this is not changing anything.

I am still a bit puzzled, though, as I cannot see such a run-and-forget guarantee in the `asyncio` sources, **so this behavior might be accidental**.  And nowhere in the docs I found any guarantee, that calling `asyncio.gather()` instead of `await asyncio.gather()` is the correct way to run things in the background.

The only thing I found in the code, which is certainly not accidental, is that following pattern indeed runs a coroutine in background, safely:

    asyncio.Task(coro())

So there is no need to keep the task object around nor to clean things up afterwards.  From the source of `class Task`:

https://github.com/python/asyncio/blob/a791d884b9110b59bada18c2dddd399e6adc40fd/asyncio/tasks.py#L69-L78

```
    def __init__(self, coro, *, loop=None):
        assert coroutines.iscoroutine(coro), repr(coro)
        super().__init__(loop=loop)
        if self._source_traceback:
            del self._source_traceback[-1]
        self._coro = coro
        self._fut_waiter = None
        self._must_cancel = False
        self._loop.call_soon(self._step)
        self.__class__._all_tasks.add(self)
```

This `self.__class__._all_tasks.add(self)` keeps the Task-object alive even if you throw it away afterwards.

**Why can't I find this essential detail prominently presented in the Docs anywhere?**

TL;DR

As `asyncio.Task()` needs a coroutine, but `asyncio.gather()` does not return one, the probably right solution is just to leave away the `await`

**But I am not sure.**

HazenBabcock: I believe that you are correct that using `await` makes `broadcast()` wait until `asyncio.gather()` finishes. This might not actually lead to a hang if there is additional buffering occurring in `WebSocketCommonProtocol.send()` or at a lower level in the OS. Eventually you'd hit `write_limit` and fail either way?

Anyway, happy to fix the example if you prefer it without the `await`.

hilbix: I do not have any powers here (and I do not want any) ;)  My intention was to help, because for me, a pure user, this technically looks as a useful example.  For me it actually was very useful already, because it made me look into the asyncio code and learn a bit better about it.  So when the `await` really is removed, there should be a good comment about why it is missing there.  **But I cannot decide if `await` shall be removed or not.**  And as I am not completely sure, others must decide anyway.

aaugustin:
Thanks for your contribution! @hilbix had a good point; this led to a very long discussion in #653, which I summarized in #986. So this PR is essentially superseded by #986.

PR #720, opened 2020-01-16
Extended logging information in client/server protocols

lburg:
While debugging some of our websocket services at work, we found some log messages to be lacking information. We had a hard time keeping track of multiple clients reconnecting and server errors.

First commit displays the state value when asserting we're in a given state while opening connection or pinging.

Then, add a custom `LoggerAdapter` to include IP address of clients, as well as a way to distinguish connections from the same remote host. We chose to display the `id` of a protocol object, but maybe there's something a bit easier to read?

Now that the `LoggerAdapter` is here, we could also move in the showing of `self.side`, but I don't know how to elegantly choose between the various separators used (`x`, `<`, `>`, `!`, `-`).

### DIFF
--- a/src/websockets/protocol.py
+++ b/src/websockets/protocol.py
@@ -325,7 +325,7 @@ class WebSocketCommonProtocol(asyncio.Protocol):
 
         """
         # 4.1. The WebSocket Connection is Established.
-        assert self.state is State.CONNECTING
+        assert self.state is State.CONNECTING, "wrong state %r" % self.state
         self.state = State.OPEN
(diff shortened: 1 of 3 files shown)


aaugustin:
I agree that logging is far from optimal. I adopted some cute conventions for my own use when I started the library and never took the time to do something serious.

I'll check out your proposal. I recently investigated structured logging options in Python. I'd like to do something that works well with logging plain old strings and also support more advanced approaches.

aaugustin:
Thanks again Lois. I'm going to follow up in #960. I will reuse your ideas.

PR #729, opened 2020-02-28
Fix typo in example/counter.py

akgnah:
Fixed an error of logging.error


aaugustin:
The right fix is likely to replace `{}` by `%s`.

You should let logging do the interpolation.

akgnah:
It has been modified.

Thanks for your guidance.

aaugustin:
Thanks for fixing this! (And sorry for not merging the PR earlier.)

PR #738, opened 2020-03-28
Use proper stack level in deprecation warnings

jack1142:
Changed the deprecation warnings to use `stacklevel=2` so that the warning will refer to the caller - the warnings are about some code calling constructor with deprecated kwarg or about some code calling deprecated property so it makes more sense to refer to the code that calls them.

### DIFF
--- a/src/websockets/client.py
+++ b/src/websockets/client.py
@@ -397,7 +397,9 @@ class Connect:
         if timeout is None:
             timeout = 10
         else:
-            warnings.warn("rename timeout to close_timeout", DeprecationWarning)
+            warnings.warn(
+                "rename timeout to close_timeout", DeprecationWarning, stacklevel=2
+            )
         # If both are specified, timeout is ignored.
         if close_timeout is None:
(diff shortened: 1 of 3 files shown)


aaugustin:
`stacklevel=2` seems mostly correct in `client.py` (renamed to `asyncio_client.py` since you submitted this PR). You'd need `stacklevel=3` when `unix_connect`, since there's an extra call (`unix_connect` -> `connect`).

The same holds for `Serve` in `server.py`.

However, `stacklevel=2` is most likely incorrect for `WebSocketServerProtocol` in `server.py`. The protocol factory is invoked by asyncio. I'm not quite sure what the call stack looks like, but it's certainly bigger than a single call.

As you can see, there's almost never a universally correct `stacklevel`, and this is why I usually don't bother setting one. Run with -Werror if you want the stack trace.

----

Besides, the intended use case for stacklevel — described in https://docs.python.org/3/library/warnings.html#warnings.warn — isn't what you're attempting to use it for.

PR #744, opened 2020-04-08
support request lines of 4107 bytes

### LINKED ISSUE #743: Consider accepting 4107 bytes for request maximum accepted line size
The code responsible for parsing a line from the request [raises `SecurityError("line too long")`](https://github.com/aaugustin/websockets/blob/139085fe2624192a5a6c72b1e5db211dcec6ced1/src/websockets/http.py#L216) when the size is greater than [`MAX_LINE = 4096`](https://github.com/aaugustin/websockets/blob/139085fe2624192a5a6c72b1e5db211dcec6ced1/src/websockets/http.py#L37)

Please consider increasing `MAX_LINE` to `4107` bytes.

# Rationale 

While the HTTP protocol does not specify limits for [request-line size](https://tools.ietf.org/html/rfc7230#section-3.1.1), [header size](https://tools.ietf.org/html/rfc7230#section-3.2.5) or [cookie size](https://tools.ietf.org/html/rfc2965#section-5.3), server implementations do (for security reasons). User agents also implement limits for the maximum size of a request-line, header, cookie, ...

So how to choose a reasonable limit for the maximum size a server implementation could enforce when parsing a requests ? We must consider what maximum sizes are enforced by popular recent user agents (let say Firefox and Chrome). 

It is safe to assume that the most frequently oversized line in a HTTP request is the one for the cookie header, since website developers tend to abuse this storage facility. On the [browsercookielimits](http://browsercookielimits.squawky.net/) website, we can discover that cookie value limits for popular browsers are : 
- Chrome : 4096 bytes
- Firefox : 4097 bytes

If we want to accept any requests made from those browsers even when they include a cookie reaching the user agent limit, we on the server side must consider accepting a request line of **4107 bytes** : 

```
cookie: xxxxxxxxxxxxxxxxxxxxxxxxxxxxxxx\r\n
\__ 8 _/\___________ 4097 ____________/\_2/    = 4107
```

thomasleveil:
fix #743

avoid sending a `HTTP 400` response when popular browsers send a request with cookies maxing up the user-agent limit

### DIFF
--- a/.appveyor.yml
+++ b/.appveyor.yml
@@ -6,7 +6,7 @@ skip_branch_with_pr: true
 
 environment:
 # websockets only works on Python >= 3.6.
-  CIBW_SKIP: cp27-* cp33-* cp34-* cp35-*
+  CIBW_SKIP: cp27-* cp33-* cp34-* cp35-* pp27-*
   CIBW_TEST_COMMAND: python -W default -m unittest
   WEBSOCKETS_TESTS_TIMEOUT_FACTOR: 100
 
--- a/src/websockets/http.py
+++ b/src/websockets/http.py
@@ -34,7 +34,7 @@ __all__ = [
 ]
 
 MAX_HEADERS = 256
-MAX_LINE = 4096
+MAX_LINE = 4107
 
 PYTHON_VERSION = "{}.{}".format(*sys.version_info)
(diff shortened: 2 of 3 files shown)


aaugustin:
I adapted your commit to recent changes in master and merged it directly.

aaugustin:
Thanks!

PR #422, opened 2018-06-02
Add support for HTTP(S) proxies to connect()

aaugustin:
Fix #364.

PR #751, opened 2020-04-24
HTTP proxy support reworked

### LINKED ISSUE #364: Add support for HTTP(S) proxies to connect()
how to make the websocket connection go through with http proxy?

andrewshadura:
Hi,

I needed HTTP support in websockets, so I took an existing PR #422 and ported it to the latest version of the main code. Since `start_tls` has been available for a while now, the hack with upgrading HTTP to HTTPS is no longer needed.

My original implementation was a [separate module](https://gitlab.apertis.org/infrastructure/obs-proxy/-/blob/master/obs_proxy/proxied_ws.py) overriding things and providing proxying support on top of the unmodified websockets library. I have tested that one quite a lot and then ported and integrated it back. I have not written any tests since I’m not quite sure how to approach this.

I have changed the implementation of `ProxyURI` to allow a trailing slash in the proxy URI. While this is non-standard, it seems to be widely supported and many users are unaware that the trailing slash is not supposed to be used in the proxy URIs.

What doesn’t work yet:
 * [ ] I have not yet touched redirects code; most likely redirects will try to connect bypassing the proxy
 * [ ] `host` and `port` keyword parameters change the host and port used to connect to the proxy, not the target
 * [ ] I was unable to get rid of one mypy error around `proxy_uri` parsing
 * [ ] I ran the code through black but didn’t agree with some changes it proposed
 * [ ] As mentioned above, some tests need to be written.

This fixes #364, supersedes and closes #422.


andrewshadura:
Oh well, last minute fixes without running tests — I should have not done that :slightly_smiling_face:

aaugustin:
Preliminary note -- Python 3.6 is officially supported until 2021. It feels a bit early to drop support for this version. Perhaps we could make HTTP proxy support a Python 3.7+ only feature, while preserving compatibility with Python 3.6 for everything else.

andrewshadura:
On Tue, Apr 28, 2020 at 07:04:53AM -0700, Aymeric Augustin wrote:
>    Preliminary note -- Python 3.6 is officially supported until 2021. It
>    feels a bit early to drop support for this version. Perhaps we could make
>    HTTP proxy support a Python 3.7+ only feature, while preserving
>    compatibility with Python 3.6 for everything else.

Sure. I can submit a revised pull request. Would you prefer `connect()` to
throw an exception when accessing an `https://` URI behind a proxy when
`start_tls` is not available or simply let it fail further down the line
while documenting it thoroughly?

-- 
Cheers,
  Andrej

aaugustin:
Perhaps it's best if I spend more time reviewing before deciding on the next steps.

lambdalisue:
Great job @andrewshadura and @aaugustin! I'm looking froward to this. Is there anything I can help with?

amatai:
This is great. I am wanting to move from autobahn for its lack of proxy w/ asyncio support.
Can I help in this PR to get it part of release?

andrewshadura:
Unfortunately, there were some upstream changes which most likely conflict with my PR, so I guess it will require some significant rework once again.

aaugustin:
Support for Python 3.6 was dropped, which lifts the biggest blocker to this.

However, in the meantime, the library has undergone a major refactoring. I'm writing a "new" implementation on top of a Sans-I/O core, which will make it vastly easier to implement this feature. I'm increasingly reluctant to add features to what has become the "legacy" implementation because this increases the size of the target. Of course, all this doesn't really help you in the short term :-/

xiaocairush:
Thanks @andrewshadura I have changed code in my local repo of version 9.1 and it works for me! There is very little conflict.

HMaker:
@andrewshadura is it broken? I tried now with websockets 10.1 and Python 3.8, the SSL handshake always fails with `decode_error` alert from server. Looking at the src code I can't find the reason, it's strange since the bug is on the SSL handshake, I am wondering if any trash is left on the transport when `loop.start_tls()` is called.

robsmith11:
@xiaocairush are you able to share the changes you made to get it working?

andrewshadura:
While we have deployed this code for the specific feature the customer has requested, the feature was well-tested and worked, they never enabled it in production, and will unlikely to do so short-term. Meanwhile we pinned `websockets` to version 8 to allow the [external implementation](https://gitlab.apertis.org/infrastructure/obs-proxy/-/blob/master/obs_proxy/proxied_ws.py) to work, but I don’t intend to work on this PR any further unless the customer actually starts using it.

Given the amount of refactoring that happened in the project since this was submitted, I think a new PR with this functionality re-implemented once again in case of demand should be more appropriate 🙂

PR #763, opened 2020-05-18
fix-#762: future proof asyncio.wait usage

### LINKED ISSUE #762: DeprecationWarning when running the tests on py3.9
`python -W default -m unittest
................................./Users/db/venvs/websockets39/lib/python3.9/site-packages/websockets-8.1-py3.9-macosx-10.15-x86_64.egg/websockets/server.py:716: DeprecationWarning: The explicit passing of coroutine objects to asyncio.wait() is deprecated since Python 3.8, and scheduled for removal in Python 3.11.
  await asyncio.wait(
../Users/db/venvs/websockets39/lib/python3.9/site-packages/websockets-8.1-py3.9-macosx-10.15-x86_64.egg/websockets/server.py:716: DeprecationWarning: The explicit passing of coroutine objects to asyncio.wait() is deprecated since Python 3.8, and scheduled for removal in Python 3.11.
  await asyncio.wait(
...../Users/db/venvs/websockets39/lib/python3.9/site-packages/websockets-8.1-py3.9-macosx-10.15-x86_64.egg/websockets/server.py:716: DeprecationWarning: The explicit passing of coroutine objects to asyncio.wait() is deprecated since Python 3.8, and scheduled for removal in Python 3.11.
  await asyncio.wait(
........................................../Users/db/venvs/websockets39/lib/python3.9/site-packages/websockets-8.1-py3.9-macosx-10.15-x86_64.egg/websockets/server.py:716: DeprecationWarning: The explicit passing of coroutine objects to asyncio.wait() is deprecated since Python 3.8, and scheduled for removal in Python 3.11.
  await asyncio.wait(
./Users/db/venvs/websockets39/lib/python3.9/site-packages/websockets-8.1-py3.9-macosx-10.15-x86_64.egg/websockets/server.py:716: DeprecationWarning: The explicit passing of coroutine objects to asyncio.wait() is deprecated since Python 3.8, and scheduled for removal in Python 3.11.
  await asyncio.wait(
.....`

DavidBord:
Fixes #762

### DIFF
--- a/src/websockets/server.py
+++ b/src/websockets/server.py
@@ -28,7 +28,6 @@ from typing import (
     cast,
 )
 
-from .compatibility import asyncio_create_task
 from .exceptions import (
     AbortHandshake,
     InvalidHandshake,
@@ -716,7 +715,7 @@ class WebSocketServer:
         if self.websockets:
             await asyncio.wait(
                 [
-                    asyncio_create_task(websocket.close(1001))
+                    asyncio.ensure_future(websocket.close(1001))
                     for websocket in self.websockets
                 ],
                 loop=self.loop if sys.version_info[:2] < (3, 8) else None,
--- a/src/websockets/server.py
+++ b/src/websockets/server.py
@@ -28,6 +28,7 @@ from typing import (
     cast,
 )
 
+from .compatibility import asyncio_create_task
 from .exceptions import (
     AbortHandshake,
     InvalidHandshake,
@@ -715,9 +716,7 @@ class WebSocketServer:
         if self.websockets:
             await asyncio.wait(
                 [
-                    asyncio.ensure_future(websocket.close(1001))
-                    if sys.version_info[:2] < (3, 7)
-                    else asyncio.create_task(websocket.close(1001))
+                    asyncio_create_task(websocket.close(1001))
                     for websocket in self.websockets
                 ],
                 loop=self.loop if sys.version_info[:2] < (3, 8) else None,
--- a/src/websockets/server.py
+++ b/src/websockets/server.py
@@ -715,7 +715,9 @@ class WebSocketServer:
         if self.websockets:
             await asyncio.wait(
                 [
-                    asyncio.create_task(websocket.close(1001))
+                    websocket.close(1001)
+                    if sys.version_info[:2] < (3, 8)
+                    else asyncio.create_task(websocket.close(1001))
                     for websocket in self.websockets
                 ],
                 loop=self.loop if sys.version_info[:2] < (3, 8) else None,
--- a/src/websockets/server.py
+++ b/src/websockets/server.py
@@ -714,7 +714,10 @@ class WebSocketServer:
         # asyncio.wait doesn't accept an empty first argument
         if self.websockets:
             await asyncio.wait(
-                [websocket.close(1001) for websocket in self.websockets],
+                [
+                    asyncio.create_task(websocket.close(1001))
+                    for websocket in self.websockets
+                ],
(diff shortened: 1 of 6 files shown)


src/websockets/server.py
```diff
@@ -714,7 +714,10 @@ async def _close(self) -> None:
         # asyncio.wait doesn't accept an empty first argument
         if self.websockets:
             await asyncio.wait(
-                [websocket.close(1001) for websocket in self.websockets],
+                [
+                    asyncio.create_task(websocket.close(1001))
```
cjerdonek, line 718: Note that `create_task()` was added in Python 3.7, and it looks like websockets will continue to support 3.6 in the next release: https://docs.python.org/3/library/asyncio-task.html#asyncio.create_task

src/websockets/server.py
```diff
@@ -715,7 +715,9 @@ async def _close(self) -> None:
         if self.websockets:
             await asyncio.wait(
                 [
-                    asyncio.create_task(websocket.close(1001))
+                    websocket.close(1001)
```
cjerdonek, line 718: For consistency, it might be better to call `ensure_future()` for 3.6: https://docs.python.org/3/library/asyncio-task.html#asyncio.create_task
(`create_task()` is basically the newer version of `ensure_future()`.)

cjerdonek: Or just always use `ensure_future()`, since this library maintains compatibility with 3.6.

src/websockets/server.py
```diff
@@ -714,7 +714,12 @@ async def _close(self) -> None:
         # asyncio.wait doesn't accept an empty first argument
         if self.websockets:
             await asyncio.wait(
-                [websocket.close(1001) for websocket in self.websockets],
+                [
+                    asyncio.ensure_future(websocket.close(1001))
+                    if sys.version_info[:2] < (3, 7)
+                    else asyncio.create_task(websocket.close(1001))
+                    for websocket in self.websockets
+                ],
```
aaugustin, line 722: Let's not add a one-off implementation of this compatibility requirement just in one place.

When we want to support Python 3.9, let's add a compatibility  function, similar to what db25f49496343bb6aacbe31994da83a5470dc67c removed.

cjerdonek: Strictly speaking, for the purposes of this issue, it's not really necessary to use `asyncio.create_task()` (since `ensure_future()` hasn't actually been deprecated). So this issue could also be addressed simply by using `ensure_future()` and not adding a compatibility layer.

In terms of a compatibility layer, I'm not sure if having the ability to use `create_task()` in higher versions outweighs the maintenance burden of the compat function, as well introducing different code paths (and potentially different behavior) for different versions.

aaugustin: Indeed — I didn't realize that ensure_future still worked on Python 3.9.

The Python docs only say that create_task is "higher-level" than ensure_future. I don't expect there will be a noticeable difference for our usage.

Like Chris, I'd be inclined to stick with ensure_future until all versions of Python supported by websockets support the create_task API, then switch from ensure_future to create_task.

cjerdonek: 👍 The other relevant statement in the docs is in the [`ensure_future()`](https://docs.python.org/3/library/asyncio-future.html#asyncio.ensure_future) docs:

> **Important:** See also the create_task() function which is the preferred way for creating new Tasks.

But nevertheless, like you suggest, I'm not sure it confers any particular advantage..

aaugustin:
I (mostly) fixed CI issues in master, then rebased, squashed and merged your PR (see #766). Thanks for your contribution!

PR #676, opened 2019-10-06
Provide a Sans-I/O interface

aaugustin:
Fix #466.

PR #779, opened 2020-06-13
Refactor header exception logic into `get_safe` method

cool-RR:
While doing the `raise from` fixes, I noticed that this logic can be refactored. Let's see whether the tests pass, and whether you find this code more elegant than the existing code :)

### DIFF
--- a/src/websockets/http.py
+++ b/src/websockets/http.py
@@ -292,6 +292,16 @@ class Headers(MutableMapping[str, str]):
         copy._list = self._list.copy()
         return copy
 
+    def get_safe(self, header: str) -> str:
+        from .exceptions import InvalidHeader
+
+        try:
+            return self[header]
+        except KeyError as exc:
+            raise InvalidHeader(header) from exc
+        except MultipleValuesError as exc:
+            raise InvalidHeader(header, f"more than one {header} header found") from exc
+
     # Collection methods
 
     def __contains__(self, key: object) -> bool:
--- a/src/websockets/handshake.py
+++ b/src/websockets/handshake.py
@@ -94,28 +94,28 @@ def check_request(headers: Headers) -> str:
 
     try:
         s_w_key = headers["Sec-WebSocket-Key"]
-    except KeyError:
-        raise InvalidHeader("Sec-WebSocket-Key")
-    except MultipleValuesError:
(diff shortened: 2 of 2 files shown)


src/websockets/http.py
```diff
@@ -292,6 +292,16 @@ def copy(self) -> "Headers":
         copy._list = self._list.copy()
         return copy
 
+    def get_safe(self, header: str) -> str:
+        from .exceptions import InvalidHeader
```
aaugustin, line 296: I guess you've had to do this to avoid an import cycle. This is a very common issue since I implemented static type checking with mypy.

The `websockets.exceptions` module is imported at the bottom at `websockets.http`, so you remove this import use `websockets.exceptions.InvalidHeader`.

It's ugly but I'd rather have a module-level import at the bottom of a module than scattered imports inside functions.

But if you're moving the code to websockets.handshake, the problem goes away.

src/websockets/http.py
```diff
@@ -292,6 +292,16 @@ def copy(self) -> "Headers":
         copy._list = self._list.copy()
         return copy
 
+    def get_safe(self, header: str) -> str:
```
aaugustin, line 295: The only difference between `get_safe` and `__getitem__` is that `get_safe` translates `KeyError` or `MultipleValuesError` into `InvalidHeader`. Having two methods that do the same thing in the same class is weird.

`websockets.http.Headers` is intended as a low-level utility that only knows about HTTP. This is why it uses `websockets.http.MultipleValuesError` instead of one of the websockets-specific exceptions in `websockets.exceptions`.

To factor out this behavior, I'd write a utility function in `websockets.handshake`. The job of this function would be to raise an exception from `websockets.exceptions` and to provide a good error message when the WebSocket protocol specifies that a given header must be provided exactly once, but it's there zero or multiple times.

You can call it `get_unique_header` for example.

aaugustin:
Also, I resumed working on #676, and I may ask you to rebase on top of #676 once it's done rather than the other way round, because the changes are much larger over there than here.

aaugustin:
I'm going to close this for now — we lost momentum — but I'm open to a PR implementing https://github.com/aaugustin/websockets/pull/779/files#r439800454

PR #785, opened 2020-06-19
Moved the metadata into setup.cfg

KOLANICH:


aaugustin:
I appreciate your effort to contribute. Unfortunately, it's quite hard to review this kind of pull request (besides CI that shows that it doesn't work).

If I want to update the packaging system, I'll need to experiment to make sure it still does everything it needs to do. It's easier to follow my own path there than to start from where someone else ended and reverse-engineer the changes.

If you're only moving configuration from setup.py to setup.py + setup.cfg, with no functional changes, I don't think this would be an improvement. I'd rather keep everything in one place.

If I could get rid of setup.py, like I did in many of my other Python projects, I would.

KOLANICH:
>If I could get rid of setup.py, like I did in many of my other Python projects, I would.

`pyproject.toml` full support for setuptools (I guess they gonna allow non-`setup.py` mode after full `pyproject.toml` support will be implemented) is not yet here, but there is a SIG making a PEP about it privately.

aaugustin:
Yes, we'll get there eventually :-)

Issue #782, opened 2020-06-17
Logging with contextvars


vedran6:
`asyncio` has native support for[ context variables](https://docs.python.org/3/library/contextvars.html#module-contextvars). For scenarios where many websocket connections are being run in the same event loop, it would be extremely useful to be able to add some context to log entries to differentiate which task is generating which log entry.

When logging, the library could use a logging adapter expecting a contextvar called `websockets_context`, and as long each `asyncio` task sets the variable, its log entries could embed that.

For example, an `asyncio` task setting that contextvar to `21` could result in an entry like:

`DEBUG websockets.protocol client [21] < Frame(fin=True, opcode=1,...`

Would there be any interest in adding this feature or in merging such a PR?

aaugustin:
Currently, logging in websockets is mostly useful for developing the library itself. I'd like to improve it but haven't found time yet.

I'm interested at looking at a proof of concept, just to see how invasive the changes would be.

Note that websockets still supports Python 3.6, but it doesn't matter if the PoC requires 3.7, we can look at this later.

vedran6:
I'll make a PoC PR then. Let's discuss afterward.

aaugustin:
A good first step would be to log through a `self.logger` in protocol instances. Then it's easier to override.

aaugustin:
See #786 for why I don't want to hardcode this particular approach. #960 aims at making it possible.

PR #786, opened 2020-06-20
Add contextvar logging

vedran6:
As discussed in #782, this would enable asyncio contextvars logging.

A task would set its context by invoking `websockets.utils.set_logging_contextvar("context...")`.

### DIFF
--- a/src/websockets/client.py
+++ b/src/websockets/client.py
@@ -34,11 +34,12 @@ from .http import USER_AGENT, Headers, HeadersLike, read_response
 from .protocol import WebSocketCommonProtocol
 from .typing import ExtensionHeader, Origin, Subprotocol
 from .uri import WebSocketURI, parse_uri
+from .utils import ContextLoggerAdapter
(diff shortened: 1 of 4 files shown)


aaugustin:
In the context of #676, I was thinking it could be even more flexible to add a `logger` argument to `serve` and `connect`. Then you could pass any logger you want.

That would solve your issue, wouldn't it?

vedran6:
Yes I suppose it would be cleaner than my current solution where I'm loading the module, grabbing the logger, wrapping it in my own adapter and injecting it back into the module using `setattr`.

aaugustin:
OK

As a first step, I'd rather allow users to provide their own loggers. That will be the most flexible.

As a second stop, I may consider stronger integration with asyncio with contextvars. However, I'm currently working on untangling websockets from asyncio (in order to support other I/O libraries), so I would only consider this in the asyncio-specific part that remains after this refactoring.

Issue #759, opened 2020-05-13
Is there a way to disable all logging or make it optional?


DobroAlex:
Greetings. I have pretty extensive logging in my app and I don't want WebSockets to show any of its logging. Is there a proper way to disable it? 

I came up with the following solution:
There is logging setup method in the app context creation there I add the following code:
`
for logger_name in {'websockets.protocol', 'websockets.server', 'websockets.client'}:
    logging.getLogger(logger_name).disabled = True
`

But I don't think of the usage of string constants as something good and this feels totally barbarous way to disable the logging.

RemiCardona:
python logging is hierarchical, try disabling the top-most logger:
```python
logging.getLogger('websockets').disabled = True
```
should do it. If not, you could add a NullHandler and disable propagation to the root logger.

DobroAlex:
I Should have guessed myself, that's pretty good and simple, thank you

PR #787, opened 2020-06-21
Add NullHandler to root logger

grisaitis:
This is a small change and makes logging a little more user friendly. Preventing issues like https://github.com/aaugustin/websockets/issues/759 and "no handler could be found" errors. 

This practice (adding a `NullHandler` to a library's root logger) is recommended in [the docs for logging](https://docs.python.org/3/howto/logging.html#configuring-logging-for-a-library). Examples include requests ([requests/\_\_init__.py](https://github.com/psf/requests/blob/db47b9b/requests/__init__.py#L136)) and urllib3 ([urllib3/\_\_init__.py](https://github.com/urllib3/urllib3/blob/bf15395/src/urllib3/__init__.py#L45)).

### DIFF
--- a/src/websockets/__init__.py
+++ b/src/websockets/__init__.py
@@ -1,5 +1,7 @@
 # This relies on each of the submodules having an __all__ variable.
 
+import logging
+
 from .auth import *  # noqa
 from .client import *  # noqa
 from .exceptions import *  # noqa
@@ -8,6 +10,9 @@ from .server import *  # noqa
 from .typing import *  # noqa
 from .uri import *  # noqa
 from .version import version as __version__  # noqa
+
+
+logging.getLogger(__name__).addHandler(logging.NullHandler())
 
 
 __all__ = [


aaugustin:
The docs say:

> If for some reason you don’t want these messages printed in the absence of any logging configuration, you can attach a do-nothing handler to the top-level logger for your library.

"If for some reason" isn't quite "recommended" :-)

This could be the the right tradeoff, though: people who don't care about logging won't get any; people who care about logging will have more work to configure a handler.

grisaitis:
No prob of course :) just wanted to bring up the option. Feel free to close.

aaugustin:
I'm going to keep this open until I have time to do a bit of research.

I'd like to check how common this is (beyond the examples you gave) and if there are no downsides (e.g. places where I added logging to make common issues easier to diagnose).

aaugustin:
The change suggested here means that log messages at level warning or above wouldn't get displayed anymore by default. I audited these messages.

I consider useful messages that will help people resolve common problems. For example, when a connection handler in the server fails, due to inversion of control, the exception doesn't bubble up to the user's code. It's useful to log it to the terminal. Currently websockets relies on *not* having a NullHandler to surface this information.

**logger.error**

- websockets.connection: unexpected errors in parser => probably useful, eventually — this code isn't used anywhere yet
- websockets.asyncio_server: Error in connection handler => useful
- websockets.protocol: Error in data transfer => not directly useful because it isn't supposed to happen; that said, if people see it, I'll likely consider it a bug in websockets, so it's in my own interest to keep it :-)

**logger.warning**

- websockets.asyncio_server: Error in opening handshake => useful
- websockets.asyncio_server: Error in closing handshake => useful
- websockets.protocol: Unexpected exception in keepalive ping task => same comment as for "Error in data transfer"
- websockets.server: Error in opening handshake => useful, eventually — this code isn't used anywhere yet

It would be a regression not to show any of these by default, and therefore, I'm not going to follow up.

Thanks for the suggestion anyway!

Issue #784, opened 2020-06-18
HTTP Digest authentication


wom-bat:
Hi folks,
  How can I create a websocket client where the server requires digest authentication?

aaugustin:
If I remember correctly Digest Authentication requires two HTTP requests: one to get the challenge from the server, one to send the response and open the websocket connection.

If I had to do this, here's what I'd try:

1. make the first HTTP request and get the challenge from the response; this will likely easier to do with another library than websockets, as `websockets.connect` raises an exception if it fails to establish a WebSocket connection;
2. calculate the proper authentication header and send it with the `extra_headers` argument of `websockets.connect`.

It would be interesting to build this into websockets. Since we already handle basic auth and http redirects, I think we have all the pieces we need.

wom-bat:
It's a little more complex than that.  The `Authorization`: header to be sent changes on each request, so it can't be calculated just once.  What's more the server can issue a new challenge at any time.

I don't know how that fits into websockets.  It may be that we can do it just once and then after the upgrade there's no need.  I'll do some experiments on Monday.

aaugustin:
In a WebSocket connection you send HTTP headers only once. After you've upgraded from a HTTP to a HTTP connection you're fine.

apteronal:
Two options have been discussed for adding digest authentication support:
1. Using a separate library for the digest authentication.
2. Implementing digest authentication within this library.

Option 1 is disfavored by the maintainer because it adds a dependency to websockets which currently has none.  Nicolas-Feude has done some great work in #1111 toward option 2.  However, looking at the code in #1111, it would seem to add a lot of complexity to this module, and duplicates code from requests which may diverge over time.  It also seems fraught to re-implement security-related code.

I'm wondering if another option would be to use a library for the digest authentication but as an *optional* dependency.   When digest authentication is required, websockets could output a message saying that the dependency needs to be installed.  Digest authentication is an uncommon need, so it may be reasonable for the user to have to install an extra package to support it.  Not an experienced Python developer so not sure whether this is a good idea, but wanted to suggest it just in case.

aaugustin:
Unfortunately #1111 isn't on the right track for the following reasons so I'm closing it.

First, it targets the legacy implementation. However, I'm not adding new features to that implementation. The feature should target the new asyncio and threading implementation.

Second, the fundamental design problem lies in how we handle HTTP sequences prior to the WebSocket handshake. Currently, websockets supports only one request and one response while Digest Authentication requires two.

[...]

PR #788, opened 2020-06-22
Add Digest Authentication to client

wom-bat:
Uses the python3-digest package to decode the HTTP digest authorisation challenge
and create an appropriate extra header to authenticate.

This addresses issue #784 -- it's really a straw-man PR, as a start at how one might do it. I'm sure there are better ways...

### DIFF
--- a/src/websockets/__init__.py
+++ b/src/websockets/__init__.py
@@ -12,6 +12,7 @@ from .version import version as __version__  # noqa
 
 __all__ = [
     "AbortHandshake",
+    "AuthenticationRequest",
     "basic_auth_protocol_factory",
     "BasicAuthWebSocketServerProtocol",
     "connect",
--- a/src/websockets/client.py
+++ b/src/websockets/client.py
@@ -8,10 +8,12 @@ import collections.abc
 import functools
 import logging
 import warnings
+from python_digest import parse_digest_challenge, build_authorization_request 
 from types import TracebackType
 from typing import Any, Generator, List, Optional, Sequence, Tuple, Type, cast
 
 from .exceptions import (
+    AuthenticationRequest,
     InvalidHandshake,
     InvalidHeader,
     InvalidMessage,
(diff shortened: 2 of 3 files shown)


aaugustin:
Thanks for the PR. The approach looks correct.

Currently websockets has no external dependencies. I'll take a look at how hard it is to reproduce the digest auth algorithm. Probably not very hard; using a dependency was a good idea for this PoC.

I'm half-way through a gigantic refactor (#676) so I'd like to avoid creating conflicts for a few more days. That said, I'm optimistic that we'll make this work :-)

lex-talionis-emmainternational:
Is there an example on how to use this option? I took a look at the diff for the Commit, but I don't see anything obvious like a change in the ws.connect() method.  Are you passing in an extra_header or is there something I'm missing that turns on Digest auth?  If I can get it working I'm happy to add an example in the vein of [basic_auth_client.py](https://github.com/aaugustin/websockets/blob/master/example/basic_auth_client.py)

wom-bat:
You just connect:

`uri='ws://user:password@host/api'`
websocket = await connect(uri)`

lex-talionis-emmainternational:
Thank you!  I tracked the issue down to the server I'm connecting to - it does not form 401's correctly when given a basic auth header.  Instead, it terminates the TCP connection abruptly, which was stopping your code from detecting the auth type correctly.  This behavior is against the HTTP standard but alas, I'm stuck with this server.  I'd be happy to send through packet captures, curl logs, or the logs from urllib3/websockets if they would help anyone.  Not sure it is worth going out of the way to support a device that isn't conforming to standards, but some way to force digest auth would probably be all it takes.

apteronal:
@wom-bat, thank you for this PR.  It was a huge help and enabled me to use digest authentication for my application.

In my case, I had to hack python3-digest to not require the 'opaque' field because the server my application connects to does not send it.  It looked like python3-digest is not actively maintained so I didn't try submitting an issue.

The PR no longer merged for me due to changes to websockets.  Attached is a patch that re-implements it on the latest websockets, in case it's useful to someone:
[788_2023-04-12_patch.txt](https://github.com/python-websockets/websockets/files/11217963/788_2023-04-12_patch.txt)

(GitHub won't allow me to upload it with the .patch extension.  If anyone could tell me what I'm doing wrong I'd appreciate it.)

aaugustin:
Hello, sorry this wasn't a priority for me back and I'm going to close it for [the reasons described in the issue](https://github.com/python-websockets/websockets/issues/784#issuecomment-2629303214) (for future reference).

PR #799, opened 2020-07-19
Modify SSLContext's protocol

smartswdeveloper:
Changed protocol to allow that was not working in all the situations.
In particular with mkcert certificates and uvicorn

### DIFF
--- a/example/secure_client.py
+++ b/example/secure_client.py
@@ -7,7 +7,7 @@ import pathlib
 import ssl
 import websockets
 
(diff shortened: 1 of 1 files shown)


aaugustin:
I'm sorry but I don't want to make the change you're proposing for two reasons:

- `PROTOCOL_TLSv1_2` is deprecated on all versions of Python that websockets currently supports
- You define "is working" as "allows connections to establish without any regards for security". I take issue with this definition because the point of TLS is to provide security. If you want to disable `check_hostname`, go ahead — that's you choice. In the documentation, I'd rather demonstrate the secure version.

PR #815, opened 2020-08-19
Add aarch64 wheel build

janaknat:
TravisCI allows creation of arm64 wheels.

Build log: https://travis-ci.com/github/janaknat/websockets/jobs/375133111

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -13,6 +13,13 @@ matrix:
       python: "3.7"
       services:
         - docker
+    - language: python
+      dist: xenial
+      sudo: required
+      python: "3.7"
+      arch: arm64
+      services:
+        - docker
     - os: osx
       osx_image: xcode8.3
 


aaugustin:
Oh, cool. Thanks!

janaknat:
@aaugustin , there is a PR to fix an alignment issue in aarch64 wheels generated by manylinux.

https://github.com/pypa/manylinux/pull/739

The gist: using aarch64 wheels in an environment where the alignment differs from that of the host used to created said wheel will cause it to fail.

It would be best to wait for that fix to land before creating an aarch64 wheel.

AGSaidi:
The fix above has landed and the updated manylinux_2014 container is now available

janaknat:
@aaugustin We don't see the wheels for aarch64 in PyPI yet. Are there any more changes to be done?

aaugustin:
The next release will produce them. Given that support for these wheels is pretty fresh and there isn't a huge demand I'd rather wait a bit before publishing them — and fielding potential issues.

janaknat:
@aaugustin Can I know when the next release will be?

aaugustin:
Well, I never really plan releases. At irregular intervals, I go through the bug tracker, handle the tickets I find interesting and doable, and make a release.

The 80€/month I make maintaining this library pay for prompt handling of any security issue, should one arise, but not much else :-)

It's pretty stable this days and doesn't require frequent maintenance.

aaugustin:
FYI #845 is the place to discuss the 9.0 release.

janaknat:
@aaugustin Any timeline for the next release?

aaugustin:
Released. aaarch wheels are on PyPI.

aaugustin:
Given the poor signals coming from Travis CI since the company was sold, I changed the process for building wheels to GitHub actions. 9.0.1 is the same as 9.0, except wheels are built with this new process. It relies on cross-compilation.

It would be super useful if you could confirm that these wheels actually work like you expect!

janaknat:
@aaugustin Thank you so much for the release. I installed websockets with pip3 on a Graviton2 instance and tested with:
```
>>>import websockets
>>>print(websockets.__version__)
9.0.1
```
Is this a good test? Do you have any test in particular that I should run?

aaugustin:
The other two things that would be useful to check:

1. Did it install from a wheel — not from source?

2. Could you run this?

```python
from websockets.speedups import apply_mask

apply_mask(b"12345687", b"\x01\x02\x03\x04")
```

janaknat:
1. It installed from the wheel:
```
$ pip3 install websockets
Collecting websockets
  Downloading websockets-9.0.1-cp38-cp38-manylinux2014_aarch64.whl (101 kB)
     |████████████████████████████████| 101 kB 13.3 MB/s 
Installing collected packages: websockets
Successfully installed websockets-9.0.1
```
2. Output:
```
$ python3
Python 3.8.5 (default, Jan 27 2021, 15:41:15) 
[GCC 9.3.0] on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> from websockets.speedups import apply_mask
>>> apply_mask(b"12345687", b"\x01\x02\x03\x04")
b'000044;3'
>>> 
```

aaugustin:
Fantastic! Thank you very much!

PR #820, opened 2020-08-27
Support for 1012, 1013 and 1014 close codes

konichuvak:
Following the motivation of https://github.com/netty/netty/pull/8658 this PR adds 3 additional close codes that would be recognized in the close frames. 
This helps with clarifying reasons for WS closures and avoids misleading `websockets.exceptions.ProtocolError: invalid status code`.

In addition, the second commit replaces a `list` with a `set` for storing close codes, which gives a nice speedup for loookups.

### DIFF
--- a/src/websockets/exceptions.py
+++ b/src/websockets/exceptions.py
@@ -68,6 +68,8 @@ class WebSocketException(Exception):
     """
 
 
+# https://www.iana.org/assignments/websocket/websocket.xhtml
+# see "WebSocket Close Code Number Registry" section for details
 CLOSE_CODES = {
     1000: "OK",
     1001: "going away",
@@ -81,6 +83,9 @@ CLOSE_CODES = {
     1009: "message too big",
     1010: "extension required",
     1011: "unexpected error",
+    1012: "service restart",
+    1013: "service overload",
+    1014: "bad gateway",
     1015: "TLS failure [internal]",
 }
 
--- a/src/websockets/frames.py
+++ b/src/websockets/frames.py
@@ -54,7 +54,8 @@ CTRL_OPCODES = OP_CLOSE, OP_PING, OP_PONG
 
 # Close code that are allowed in a close frame.
 # Using a list optimizes `code in EXTERNAL_CLOSE_CODES`.
-EXTERNAL_CLOSE_CODES = [1000, 1001, 1002, 1003, 1007, 1008, 1009, 1010, 1011]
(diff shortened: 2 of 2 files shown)


aaugustin:
Did you benchmark replacing the list with a set? I did, and I was surprised to see that the list is short enough to be faster than a set. This is why I added a comment.

konichuvak:
Could it be possible that you have included the creation of the collection itself in your test?
I was intrigued and run some benchmarks as you suggested and here are the results I got doing 1e6 iterations of various member lookups:
```
collection             999       1000       1002       1008       1014    10000000.0          0      check
---------------  ---------  ---------  ---------  ---------  ---------  ------------  ---------  ---------
<class 'tuple'>  0.188671   0.070076   0.0902588  0.117137   0.181207      0.297242   0.168082   0.208051
<class 'list'>   0.14635    0.0626006  0.0797667  0.102241   0.147922      0.287174   0.141091   0.186147
<class 'set'>    0.0664663  0.0662823  0.0673453  0.0673237  0.0667578     0.0756681  0.0667592  0.0666796
```
... and the `set` looks better on all tests except for the case when the member is a first element in order (`1000` in this case).

---

Below is the code I used for benchmarking.
```
import timeit
from collections import defaultdict

from tabulate import tabulate

collections = [
    (1000, 1001, 1002, 1003, 1007, 1008, 1009, 1010, 1011, 1012, 1013, 1014),
    [1000, 1001, 1002, 1003, 1007, 1008, 1009, 1010, 1011, 1012, 1013, 1014],
    {1000, 1001, 1002, 1003, 1007, 1008, 1009, 1010, 1011, 1012, 1013, 1014}
]

members = [999, 1000, 1002, 1008, 1014, 10e6, 0, 'check']

results = defaultdict(list)

for collection in collections:
    results['collection'].append(str(type(collection)))
    for member in members:
        perf = timeit.timeit(lambda: member in collection, number=int(1e6))
        results[member].append(perf)

print(tabulate(results, headers=['collection'] + members))
```

aaugustin:
Well I must have missed something. Benchmarking is hard!

I'll take a second look at this (docs, tests...) ASAP then merge this PR. Thanks for your contribution!

konichuvak:
Oh, thank you for such a clean and well thought-out library!

aaugustin:
Squashed and merged with minor adjustements in a58540d.

aaugustin:
Thank you!

PR #908, opened 2021-02-18
add insecure option for cli

### LINKED ISSUE #907: How to ignore certs or trust custom cert for interactive client?
I want to use the interactive client described on [this page](https://websockets.readthedocs.io/en/stable/intro.html?highlight=interactive#one-more-thing)

```
python -m websockets wss://echo.websocket.org/
```

The particular server I'm using provides a self-signed certificate.
I want to make websockets trust this certificate, or just ignore the certificate.
I want the equivalent of `-k` or `-capath` in `curl`.

Is there anything like that? I can't see any arguments in `__main__.py` for this. But do any of the underlying network libraries read from environment variables to get configuration?

I don't want to install this certificate as a root CA for my whole laptop.

I'm considering building a docker image and installing this particular self-signed cert as the root CA inside the docker image, then running from there. But that will be fiddly because docker on mac just can't handle proxies properly. Is there an easier way?

If there's not currently a way to do this easily, would you be open to a PR to add more arguments [here](https://github.com/aaugustin/websockets/blob/aa93c4ceca90a1798f86b2fc2b110a42f308d721/src/websockets/__main__.py#L173)? I'm not yet comfortable writing with asyncio, but fortunately this part of the code seems synchronous.

mdavis-xyz:
Fixes #907 

This is only for the `-k` option. I haven't added the `--capath` option yet.
I think that would involve [this](https://docs.python.org/3/library/ssl.html#ssl.SSLContext.set_default_verify_paths).

### DIFF
--- a/docs/intro.rst
+++ b/docs/intro.rst
@@ -207,3 +207,7 @@ One more thing...
 ``websockets`` provides an interactive client::
 
     $ python -m websockets wss://echo.websocket.org/
+
+To see the usage options, run
+
+    $ python -m websockets --help
--- a/src/websockets/__main__.py
+++ b/src/websockets/__main__.py
@@ -5,6 +5,7 @@ import signal
 import sys
 import threading
 from typing import Any, Set
+import ssl
 
 from .exceptions import ConnectionClosed, format_close
 from .legacy.client import connect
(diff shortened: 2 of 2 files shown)


mdavis-xyz:
Note that I haven't written any automated tests for this. I'm not fluent with asyncio, so I'm not sure the best way to do this.

I tested this manually on a server presenting a self signed cert. It connects with `-k` and refuses to connect without `-k`.

mdavis-xyz:
One of the CI tests is failing.

I cannot see the error message unless I grant CircleCI permission to read and write every repository, including writing to my private github repositories, and writing to every public repo as me. That's ridiculous. I'm not going to do that just to look at scan results for a public repo.
Please consider changing your CI provider because this is unacceptable.

mdavis-xyz:
Ok I have just added a `--cacert` option too.

aaugustin:
As explained in #907, I'm proposing that you maintain your own version of this script — and maybe publish it :-) — because this opens the gates for adding many other flags and I don't wish to take this route.

PR #925, opened 2021-03-23
work around...

lsaavedr:
Please check this work around to manage IncompleteReadError exception raised when reading data with self.reader.readexactly,

regards!
- Luis


aaugustin:
This change breaks protocol compliance, as show by failures in the test suite.

I'm not seeing what problem you're trying to solve. These exceptions are caught and handled properly in `transfer_data`, from which you copy-pasted the comment.

aaugustin:
To be clear, I'm happy to consider changes if you explain the actual problem :-)

Issue #931, opened 2021-04-24
Broadcasting same message to many clients


AmirHmZz:
When it comes to deployment in large scale , optimization becomes more important.
In some cases, websocket servers want to broadcast **one same message** to all connected clients repeatedly. Is there any way to do all necessary things with one message ( **framing** + **compressing** ) and just send the prepared message to many clients ? #932

aaugustin:
Currently websockets doesn't provide an API to do this.

This feature request comes up once in a while. Let's keep this issue open to track it.

AmirHmZz:
I've tried to add this feature in a PR,
Anyway, is there anyway to do this without this PR ?
I need `Frame.serialize()` method in `lagacy.framing.Frame` but it seems it's not accessible from outside of package.

aaugustin:
Currently, the entire logic for writing frames takes a couple dozen lines:

https://github.com/aaugustin/websockets/blob/dda3dfa992ddf6045be48c34143e4c1656dff9d4/src/websockets/legacy/protocol.py#L1078-L1091

https://github.com/aaugustin/websockets/blob/dda3dfa992ddf6045be48c34143e4c1656dff9d4/src/websockets/legacy/framing.py#L100-L124

If you're ready to use private APIs, you can do something like:

```python
frame = Frame(fin, Opcode(opcode), data)
data = frame.serialize(mask=mask, extensions=extensions)
for websocket in websockets:
    if websocket.state is State.OPEN:
        websockets.transport.write(data)
```

aaugustin:
Also, thanks for submitting a PR. I reviewed it; unfortunately I'm not convinced this is the best API for providing this functionality. I put a lot of focus on designing a nice API in websockets and it's hard to change later so I'd like to invest effort upfront to figure out the best possible solution.

AmirHmZz:
Thanks for your review and solution.
Of course I wanna try private apis , but the only problem is when I do something like :
```
from websockets.framing import Frame
Frame(True,0,b"").serialize()
```
I'll get an error :
`AttributeError: 'Frame' object has no attribute 'serialize'`
As I said it seems accessible `Frame` class from outside of package is [`websockets/frames.Frame`](https://github.com/aaugustin/websockets/blob/dda3dfa992ddf6045be48c34143e4c1656dff9d4/src/websockets/frames.py#L76) which doesn't have a method called `serialize()`. 
[...]

PR #932, opened 2021-04-25
Add ability to prepare non-fragmented messages

AmirHmZz:
With these changes, server protocol will be able to serialize ( framing + compression ) a non-fragmented message in order to send it to multiple clients. #931

### DIFF
--- a/src/websockets/legacy/protocol.py
+++ b/src/websockets/legacy/protocol.py
@@ -530,6 +530,30 @@ class WebSocketCommonProtocol(asyncio.Protocol):
 
         return message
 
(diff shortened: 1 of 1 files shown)


aaugustin:
I believe the design of this PR doesn't meet the quality standards for websockets.

Indeed, `self.serialize_frame` may use arbitrary state from `self`, including:

- what extensions were configured with what parameters
- the state of the `zlib.compressobj` compressor (unless it was disabled with the "no context takeover" parameter)

This could result in sending incorrect frames. This should be impossible with public APIs.

Further problems may arise with other extensions.

PR #934, opened 2021-05-01
Add a chat system example.

yuvallangerontheroad:
A work in progress, but it uses several asyncio building blocks in non-trivial (to me) ways.
I still need to add sharing of messages between connected clients.

### DIFF
--- a/example/chat_system/client.py
+++ b/example/chat_system/client.py
@@ -33,7 +33,6 @@ async def main():
 
     while True:
         user_input = await inputter()
-        print(user_input)
         await user_input_queue.put(user_input)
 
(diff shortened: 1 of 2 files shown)


aaugustin:
Hello! Thanks for sharing this code!

It would require quite a bit of text to be a useful part of the docs — the `examples` directory contains examples used in the docs. See https://github.com/aaugustin/websockets/commit/60e9531eb978b41deee1db8552b27faa749aa515 for the sort of documentation I have in mind.

Since this is work in progress, I can't make much of it, however I would encourage you to blog about it :-) The explanation is at least as important as the code :-)

PR #936, opened 2021-05-02
Drop Python 3.6 & support Python 3.10

### LINKED ISSUE #916: Python 3.10 support (loop parameter was removed from Lock())
Hello. I'm trying python 3.10 and it seems like websockets doesn't support it yet.

How to reproduce:
Get a working websockets server and try to run a client with python3.10:
```import asyncio
import websockets

async def hello():
    uri = "ws://localhost:8100"
    async with websockets.connect(uri) as websocket:
        await websocket.send('test')
asyncio.get_event_loop().run_until_complete(hello())
```

This code breaks with such error:
```
  File "ws_cl.py", line 8, in <module>
    asyncio.get_event_loop().run_until_complete(hello())
  File "/usr/lib/python3.10/asyncio/base_events.py", line 641, in run_until_complete
    return future.result()
  File "ws_cl.py", line 6, in hello
    async with websockets.connect(uri) as websocket:
  File "venv/lib/python3.10/site-packages/websockets/client.py", line 517, in __aenter__
    return await self
  File "venv/lib/python3.10/site-packages/websockets/client.py", line 535, in __await_impl__
    transport, protocol = await self._create_connection()
  File "/usr/lib/python3.10/asyncio/base_events.py", line 1080, in create_connection
    transport, protocol = await self._create_connection_transport(
  File "/usr/lib/python3.10/asyncio/base_events.py", line 1098, in _create_connection_transport
    protocol = protocol_factory()
  File "venv/lib/python3.10/site-packages/websockets/client.py", line 69, in __init__
    super().__init__(**kwargs)
  File "venv/lib/python3.10/site-packages/websockets/protocol.py", line 235, in __init__
    self._drain_lock = asyncio.Lock(
  File "/usr/lib/python3.10/asyncio/locks.py", line 77, in __init__
    super().__init__(loop=loop)
  File "/usr/lib/python3.10/asyncio/mixins.py", line 17, in __init__
    raise TypeError(
TypeError: As of 3.10, the *loop* parameter was removed from Lock() since it is no longer necessary
```
I've looked into websockets/protocol.py:235 and there's already exist some check of python version:

```
self._drain_lock = asyncio.Lock(
    loop=loop if sys.version_info[:2] < (3, 8) else None
)
```
But from what I understand, this is not enough to solve this problem.

aaugustin:
Fix #916.

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -38,7 +38,7 @@ jobs:
     runs-on: ubuntu-latest
     strategy:
       matrix:
-        python: [3.6, 3.7, 3.8, 3.9]
+        python: [3.7, 3.8, 3.9]
     steps:
     - name: Check out repository
       uses: actions/checkout@v2
--- a/.github/workflows/wheels.yml
+++ b/.github/workflows/wheels.yml
@@ -50,7 +50,7 @@ jobs:
       uses: joerick/cibuildwheel@v1.11.0
       env:
         CIBW_ARCHS_LINUX: auto aarch64
-        CIBW_BUILD: cp36-* cp37-* cp38-* cp39-*
+        CIBW_BUILD: cp37-* cp38-* cp39-*
     - name: Save wheels
       uses: actions/upload-artifact@v2
       with:
--- a/README.rst
+++ b/README.rst
@@ -125,7 +125,7 @@ Why shouldn't I use ``websockets``?
   and :rfc:`7692`: Compression Extensions for WebSocket. Its support for HTTP
   is minimal — just enough for a HTTP health check.
 * If you want to use Python 2: ``websockets`` builds upon ``asyncio`` which
-  only works on Python 3. ``websockets`` requires Python ≥ 3.6.1.
+  only works on Python 3. ``websockets`` requires Python ≥ 3.7.
 
 What else?
 ----------
--- a/README.rst
+++ b/README.rst
@@ -45,15 +45,14 @@ Here's how a client sends and receives messages:
     #!/usr/bin/env python
 
     import asyncio
-    import websockets
+    from websockets import connect
 
     async def hello(uri):
-        async with websockets.connect(uri) as websocket:
+        async with connect(uri) as websocket:
             await websocket.send("Hello world!")
             await websocket.recv()
 
-    asyncio.get_event_loop().run_until_complete(
-        hello('ws://localhost:8765'))
+    asyncio.run(hello('ws://localhost:8765'))
 
 And here's an echo server:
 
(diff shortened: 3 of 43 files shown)


adriansev:
@aaugustin Hi! I just seen this, sorry for my lack/delayed of reaction. Is it really needed the drop of 3.6 support? EL7 have python 3.6.8 and, while most if not all CERN experiments have their own software stacks, everything is still based on EL7 (and there is no newer python in softwarecollections ). Thank you!

aaugustin:
I understand your concern. My take is the following: if you're using a Python released on Dec 23rd, 2016 and EOL on Dec 23rd, 2021, then it should be OK to use a websockets released on May 1st, 2021 and EOL some time in 2021 or maybe 2022. (Since I don't do support branches, each websockets release is EOL as soon as the next release is out.) Existing versions aren't going away.

websockets tries to maintain backwards compatibility over 5 years, which means I barely started removing backwards compatibility for changes that happened in the first years of the library. I also support 5 years old Python versions. That's a pretty strong commitment as far as open source projects maintained by a single person in their free time go.

Support multiple Python versions in websockets has adds mental overhead to remember what constraints there are on the newer and older Python versions. I don't find it reasonable to support more than 4 Python versions at a time. Given that Python does deprecations over two releases, that's two (overlapping) deprecations periods, and asyncio still changes quite a bit.

That said, if there was a significant bug in 9.0.1, then I'd probably make a branch to fix it and avoid stranding Python 3.6 users on websockets 8.x. This situation hasn't come up when I dropped 3.3, 3.4, and 3.5.

adriansev:
@aaugustin Thanks a lot for info! My only concern was for bugs that we might hit and could not be applied on the old versions (but given that everything worked so well so far, and the data-path that we use within websockets will not change, i doubt that unexpected things will arise.. but well, in engineering anything that can go wrong, it will :)  ).
So, your support statement is quite titanic especially the manpower (and that "small" detail of "free time"), so i personally i'm deeply grateful for all your work!
Also, very true your statement about multiple python versions, is quite bad that there is such big mismatch between what upstream release and what distributions are willing to package and provide. 
At this moment 9.0.1 works seemingly without problems on centos7 (python 3.6.8) so everything is well :)   
Thanks a lot for all your work and support!

aaugustin:
Thanks for the feedback on 9.0.1 :-) Since I don't even use this project myself (just maintaining it for fun) I'm always shooting in the dark when I make new releases!

adriansev:
@aaugustin well, you could do like some other projects and just announce a RC tag with a RC release before the final tag (well, actually do a release branch then announce the rc tag and release, then after testing and fixes on that branch (as well as in master) you do the tag and forget about it ... but of course is a matter of preference and available time as this requires a little bit of management)
If i can get the news in time i will test and give feedback for sure :)
Thank you!

PR #943, opened 2021-05-06
add project_urls metadata

graingert:
specifically I was looking for the Changelog      Changelog=https://websockets.readthedocs.io/en/stable/changelog.html

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -3,7 +3,12 @@ python-tag = py37.py38.py39.py310
 
 [metadata]
 license_file = LICENSE
-
+project_urls =
+    Changelog=https://websockets.readthedocs.io/en/stable/changelog.html
+    Docs=https://websockets.readthedocs.io/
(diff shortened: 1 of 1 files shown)


aaugustin:
Spec here - https://packaging.python.org/guides/distributing-packages-using-setuptools/#project-urls

aaugustin:
Where does this metadata surface?

I'm a minimalist when it comes to overhead because it requires maintenance and metadata files like `setup.cfg` are overhead.

Searching for "python websockets changelog" in the almighty Google gives the result immediately, which seems a good way find it?

graingert:
It shows up in the sidebar eg https://pypi.org/project/pytest/

aaugustin:
Ah this is nice. Thanks for bringing it up. I'll make the change!

aaugustin:
For the record "standard" names/icons are here:
https://github.com/pypa/warehouse/blob/main/warehouse/templates/packaging/detail.html

PR #949, opened 2021-05-08
Setup CodeQL analysis

aaugustin:


aaugustin:
This is quite slow and not very useful.

Issue #951, opened 2021-05-11
Interactive client requires Python>=3.9.0


paskozdilar:
Running `python3 -m websockets wss://echo.websocket.org/` on Python 3.8.5 gives an uninformative error:
```
Traceback (most recent call last):                                                                                                                                                                                 
  File "/usr/lib/python3.8/runpy.py", line 194, in _run_module_as_main                                                                                                                                             
    return _run_code(code, main_globals, None,                                                                                                                                                                     
  File "/usr/lib/python3.8/runpy.py", line 87, in _run_code                                                                                                                                                        
    exec(code, run_globals)                                                                                                                                                                                        
  File "/home/$USER/websockets/src/websockets/__main__.py", line 215, in <module>                                                                                                                            
    main()                                                                                                                                                                                                         
  File "/home/$USER/websockets/src/websockets/__main__.py", line 181, in main                                                                                                                                
    async def queue_factory() -> asyncio.Queue[str]:                                                                                                                                                               
TypeError: 'type' object is not subscriptable
```
The issue is related to the [PEP 585 - Type Hinting Generics In Standard Collections](https://www.python.org/dev/peps/pep-0585/), which is introduced in Python 3.9.

This does not fit the Python version requirement specified in the [documentation](https://websockets.readthedocs.io/en/stable/intro.html#requirements).

The fix could be either:
- quoting the `asyncio.Queue[str]` on the line 181 of `__main__.py`, or
- importing `from __future__ import annotations` the backwards compatility.

NOTE: If the latter fix is chosen, I think it would be style-consistent to convert other quoted type hints (on lines 94 and 95) to type-hinted generics.

I have submitted a [PR](https://github.com/aaugustin/websockets/pull/952) with the latter fix.

aaugustin:
`from __future__ import annotations` doesn't work on Python 3.6, which is supported in 9.x (and removed in 10.x) so we have to go with quoting.

PR #952, opened 2021-05-11
Fix websockets client exception with Python<=3.9

paskozdilar:
Implements fix for #951.

### DIFF
--- a/src/websockets/__main__.py
+++ b/src/websockets/__main__.py
@@ -1,3 +1,5 @@
+from __future__ import annotations
+
 import argparse
 import asyncio
 import os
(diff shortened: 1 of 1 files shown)


aaugustin:
I chose another solution in order to fix in a 9.0.x release that this supports Python 3.6.

paskozdilar:
Understandable.
Sorry for not providing both PR's, I had a feeling it was an arbitrary choice, so I took the "newest-feature" road.
Cheers!

aaugustin:
Well I didn't know about `from __future__ import annotations`, now I do, and I can use it in the main branch where support for Python 3.6 is dropped. So thank you for that!

PR #957, opened 2021-05-15
Explain keepalives and high traffic.

### LINKED ISSUE #919: Best practices for ingesting high frequency data
Hello, let me start by saying thank you for the great work on this library. I've been obsessively going through different possible solutions to my problem the past week and the quality of this library really stands out.

Much of what I will present below has been discussed in other closed issues but not in one place and not with these problems (to my knowledge). Perhaps this record will also help others in my position. Here it goes...

I've been working with Python for data analysis for several years now and recently decided to learn how to set up systems for high-frequency message ingestion and analysis, for IoT and financial applications (not crypto!). I am still learning and, yes, I know Python might prove too slow for this use case.

So I signed up with polygon.io since their stock websocket stream would provide plenty of messages for testing (and learning). As of this writing they offer four streams (A, AM, T, Q), one can subscribe to either or all at a time, I think. I am focusing on the "A. *" stream which delivers per-second aggregates. I observed an average of ~550 msg/sec throughout the day with some occasional spikes of ~1250 msg/sec. I read that the "T. *" stream can peak at ~2M msg/sec but I haven't touched it. I am intrigued though, once I develop my skills. Each message is sending a single event as ``[{"ev":"A", "sym":"MSFT", "v":123, ...}]`` and it is stated that if the client is slow in handling the messages they will be sent with 2+ events as ``[{"ev":"A", "sym":"MSFT", "v":123, ...},{"ev":"Q", "sym":"AAPL", "v":456, ...},...]``.

Polygon has a very basic Python example ingesting data from the streams using the ``websocket_client`` library and threading. I find ``websockets`` and ``asyncio`` to be a far better solution and this was my initial implementation. I should mention that this script pushed the messages to a TimescaleDB instance, I am now working on publishing the messages to a Kafka instance.

```python
async def stream():
    async with websockets.connect(POLYGON_URI, ssl=True) as websocket:
        # authenticate
        await websocket.send(
            '{"action":"auth","params":"' + POLYGON_KEY + '"}')
        # subscribe
        await websocket.send(
            '{"action":"subscribe","params":"A.*"}')

        # Here I just want to remove the three status messages from processing
        # Perhaps there is a better way?
        greet = await websocket.recv()
        auth = await websocket.recv()
        sub = await websocket.recv()
        print(greet, auth, sub)  # <-- replace with log

        # message handling
        async for message in websocket:
            # do something with each message... (i.e. insert to DB, publish to MQ, etc.)
```

Simple, right? The connection was rock solid for the most part except for the "high trading activity" hours, namely 9:30-10:30am and 3:30-4:00pm where my connection would constantly drop with error 1006. What was particularly puzzling was that I did not notice a dramatic 2x, 3x message increase during these hours, maybe a 10-25% increase. I was constantly monitoring my computer's resources, the script was running on 6 threads (out of the available 24) and CPU usage never near 100%.

I haven't done a lot of testing yet but it appears that setting ``ping_timeout=None`` and ``max_queue`` to a high enough number keeps the connection alive throughout the day. For what it's worth, the connection seemed more stable on my personal rig than the GCP VMs that I tested (proxies, clock speed??). I'd really like to understand what's happening though, how I can optimize the process and how I can figure out what the limitations of Python are where adding more memory or CPU resources won't make a difference and moving on to a different language such as Go would be more appropriate. Is there a way to print/log and inspect the queue size as the messages are coming in? Are there any best practices for parsing messages in the async loop? Could my inserts to TimescaleDB be slowing down the process? Any advice on how to best troubleshoot?

Apologies for the question barrage, any input would be much appreciated. Again, great work with this library. Thanks!

aaugustin:
Fix #919.

### DIFF
--- a/docs/deployment.rst
+++ b/docs/deployment.rst
@@ -145,7 +145,69 @@ adapt them to your application. When you call :func:`~legacy.server.serve`:
 Furthermore, you can lower ``read_limit`` and ``write_limit`` (default:
 64 KiB) to reduce the size of buffers for incoming and outgoing data.
 
-The design document provides :ref:`more details about buffers<buffers>`.
+The design document provides :ref:`more details about buffers <buffers>`.
+
+Keep-alive and timeouts
+-----------------------
+
+Since the WebSocket protocol is intended for real-time communications over
+long-lived connections, it is desirable to ensure that connections don't
+break, and if they do, to report the problem quickly.
+
+Furthermore, WebSocket is built on top of HTTP/1.1, where connections are
+short-lived. Typically, HTTP/1.1 infrastructure closes idle connections after
+30 to 120 seconds. Proxies may terminate WebSocket connections prematurely,
+when no message was exchanged in 30 seconds.
+
+websockets implements a keepalive mechanism based on WebSocket Ping and Pong
+frames, which are designed for this purpose.
+
+It waits 20 seconds, then it sends a Ping frame, and it expects to receive the
+corresponding Pong frame within 20 seconds. Else, it considers the connection
+broken and closes it. Timings are configurable with ``ping_interval`` and
+``ping_timeout``.
+
+While WebSocket runs on top of TCP, websockets doesn't rely on TCP keepalive
+because it's disabled by default and, if enabled, the default interval is no
+less than two hours, which doesn't meet our requirements.
+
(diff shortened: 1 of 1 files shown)


aaugustin:
Rebased, modified and merged in 58ac522.

PR #961, opened 2021-05-15
Improve logging

### LINKED ISSUE #960: Improve logging
There were a bunch of discussions around logging in the past. I'm going to centralize what I want to achieve in this issue.

**Requirements**

1. websockets shouldn't enforce a particular logging library or style, because that may create difficulties for users of the library, especially in case of conflicts with other libraries. Off-the-shelf `logging` it is. No abstractions on top of `logging` allowed. Not that I like `logging`, but we'll make do with what we have.
2. Log messages should include an ID to tell which connection they relate to. This should be optional on the client side because client scripts may have only one connection. ~The ID should be user-settable.~ [EDIT: not really seeing a use case for this; we can reconsider later if we discover one.]
3. Messages should be logged via a logger that should be user-settable. This should allow users to pass additional context into the logger, or maybe to pass a logger adapter instead of a logger. If we wrap the user-provided logger into a logger adapter of own own, we should namespace our keys to prevent name conflicts.
4. Debug logs should be understandable by people who haven't memorized opcodes and close codes.

**Prior discussions**

- #720 proposed adding a logger adapter. It implemented a subset of point 3 above. However, it made more specific choices than what I'd like. I'd like to make it easy to implement that, but also offer other options.
- #782 / #786 also proposed adding a logger adapter which would add context from a contextvar. Again, I'd like to make it easy to implement that, but also offer other options.
- #765 is somewhat independent, but will be more comfortable to implement if logging is clean.

aaugustin:
Fix #960.


src/websockets/legacy/protocol.py
```diff
@@ -132,6 +132,19 @@ def __init__(
...
 
+        # Unique identifier. For logs.
+        self.id = uuid.uuid4()
+
+        # Logger or LoggerAdapter for this connection.
+        if logger is None:
+            logger = logging.getLogger("websockets.protocol")
+        # https://github.com/python/typeshed/issues/5561
+        logger = cast(logging.Logger, logger)
+        self.logger = logging.LoggerAdapter(logger, {"websocket": self})
```
graingert, line 143: this breaks uvicorn because `self.logger.level` no longer exists: https://github.com/encode/uvicorn/blob/e881ab06d590adbf3aa5c8333be2306abb50d077/uvicorn/protocols/websockets/websockets_impl.py#L79-L80

aaugustin: As a general good practice, I would recommend you use public APIs rather than private APIs.

The `level` attribute of `logging.Logger` objects is a private API: it isn't documented here: https://docs.python.org/3/library/logging.html#logging.Logger

The `getEffectiveLevel()` is the public API for doing what you want and it works identically for `logging.Logger` and `logging.LoggerAdapter` objects:


```python
>>> import logging
>>> logger = logging.getLogger()
>>> adapter = logging.LoggerAdapter(logger)
>>> logger.level  # PRIVATE API! DON'T DO THIS!
30
>>> logger.getEffectiveLevel()
30
>>> adapter.level
AttributeError: 'LoggerAdapter' object has no attribute 'level'
>>> adapter.getEffectiveLevel()
30
```

PR #978, opened 2021-06-03
Make it possibly to bypass the handshake.

aaugustin:
This can facilitate integration in other projects that bring
their own HTTP stack.


ashleysommer:
Looks perfect for Sanic's new Websockets integration.

PR #986, opened 2021-06-06
Everything about broadcast

aaugustin:
Ref #653.


AmirHmZz:
That's amazing! I'm so glad that I could be useful in this part of your great project :)

aaugustin:
If you don't want to wait for a release, you can install the main branch with:
```
pip install git+https://github.com/aaugustin/websockets.git@fa497e50
```

Issue #767, opened 2020-05-23
Some attributes not initialized in WebSocketCommonProtocol


cjerdonek:
I had some code that was working prior to 8.0 that stopped working when upgrading. When a connection ended, the code did some debug logging like logging `protocol.close_code`. Starting with 8.0, that code started raising `AttributeError` because the attribute (along with some others) aren't initialized starting with 8.0:
https://github.com/aaugustin/websockets/commit/94945fec3280c1c6d04aa4c9dc41a794f1ef1a42#diff-5e2d99fc26f0a3387bd0ff4241906a9bL225-R240
The changelog also doesn't mention that these attributes would be going away. Some of the other attributes affected are `path` and `request_headers`. Is there a reason the attributes can't all be initialized to some value like `None`, or an equivalent for the attribute's type?

aaugustin:
If you initialize them to None, type declarations become more complicated.

This mypy experiment is a complete disaster.

cjerdonek:
Are you sure? For example, what about initializing `close_code` to 0? It seems like the type checker would be just as happy with that.

aaugustin:
Yes, the type checker would be happy with that, but I wouldn't. The close code is undefined until we receive a close frame (or until we decide we'll never receive one because the connection has dropped). Usually undefined values are set to None in Python. That's what websockets used to do.

In order to avoid a bunch of `Optional[...]` I opted not to define the at all until we know the value. That was probably a mistake.

There's an easy, but ugly, way to restore backwards-compatibility: rename `close_code` to `_close_code` and add:

```python
    @property
    def close_code(self):
        return getattr(self, `_close_code`, None)
```

The alternative is to bite the bullet, use `Optional[None]` and add whatever needed for mypy to stop complaining. This may be feasible; I don't think I tried very hard.

aaugustin:
`close_reason` has the same problem and should be treated similarly to `close_code`.

Ditto for all other public attributes that are in the same situation.

I don't think we have to make changes for the private attributes.

cjerdonek:
Okay, I wasn't sure if it was the Optional you were trying to avoid. I agree mypy doesn't necessarily make things easy. I think it's probably just a lot harder when trying to do it all at once as opposed to incrementally, where you can address these things at a slower pace.

aaugustin:
[...]

PR #987, opened 2021-06-07
support lazy construction of WebSocketCommonProtocol

graingert:


src/websockets/legacy/protocol.py
```diff
@@ -216,6 +207,30 @@ def __init__(
...
         self.close_connection_task: asyncio.Task[None]
+            
+    @property
+    def loop(self) -> asyncio.AbstractEventLoop:
+        return asyncio.get_running_loop()
+    
+    @loop.setter
+    def loop(self, value: asyncio.AbstractEventLoop) -> None:
+        if value is not asyncio.get_running_loop():
+            raise AttributeError("can't set attribute)
```
graingert, line 218:
```suggestion
            raise AttributeError("can't set attribute")
```

graingert:
uvicorn constructs WebSocketCommonProtocol objects before the loop starts running

aaugustin:
I understand you have a problem and I'd like to help you resolve it.

However, your explanation of the problem is quite terse:

> uvicorn constructs WebSocketCommonProtocol objects before the loop starts running, defer retrieval of the loop until absolutely necessary (when connection_made is called)

Furthermore, the patch quite large and includes potentially backwards incompatible changes where consequences are hard to assess, like this one:

```diff
-        self._drain_lock = asyncio.Lock(**loop_if_py_lt_38(loop))
+        self._drain_lock = asyncio.Lock()
```

You're also removing some attributes that used to be set in the constructor. This will definitely cause trouble; similar changes caused #767 previously. Let's not make this mistake again.

It seems to me that you're getting burnt by the change in Python 3.10 that makes many asyncio APIs non functional unless an event loop is running. This change was quite painful for websockets as it's test suite relied heavily on doing setup, then running the event loop, then checking results.

Could we discuss further the problem you're facing, review possible solutions, and look for a design that creates fewer backwards-compatibility risks?

aaugustin:
Is this related to https://github.com/encode/uvicorn/pull/1070?

graingert:
@aaugustin yep that's the PR

aaugustin:
I see that you continue pushing code changes so I'm afraid I didn't get the right message across.

Currently this PR is headed the wrong way. I cannot merge it because it contains too many changes that I disagree with and where we didn't seek consensus on the design, even after I asked you to stop coding and start discussions. My concerns include but aren't limited to:

- taking risks with backwards compatibility in the `legacy` module, including but not limited to:
    - changing attributes of protocols
    - changing sequences of events — it's basically impossible to tell if these changes can cause side effects in review
- moving things around without a clear reasoning and a clear plan
- `if TYPE_CHECKING`

aaugustin:
Please file an issue explaining the problem you're trying to solve, so we can figure out a solution. Thank you.

PR #998, opened 2021-06-16
Add details to ConnectionClosed exceptions.

aaugustin:
Still TODO

- [x] Add API documentation of the new ConnectionClosed attributes

### DIFF
--- a/src/websockets/exceptions.py
+++ b/src/websockets/exceptions.py
@@ -90,24 +90,25 @@ class ConnectionClosed(WebSocketException):
         self.rcvd = rcvd
         self.sent = sent
         self.rcvd_then_sent = rcvd_then_sent
-        if rcvd is None:
-            if sent is None:
-                assert rcvd_then_sent is None
-                msg = "no close frame received or sent"
+
+    def __str__(self) -> str:
+        if self.rcvd is None:
+            if self.sent is None:
+                assert self.rcvd_then_sent is None
+                return "no close frame received or sent"
             else:
-                assert rcvd_then_sent is None
-                msg = f"sent {sent}; no close frame received"
+                assert self.rcvd_then_sent is None
+                return f"sent {self.sent}; no close frame received"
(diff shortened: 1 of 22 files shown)


src/websockets/exceptions.py
```diff
@@ -81,37 +81,63 @@ class ConnectionClosed(WebSocketException):
...
+        else:
+            if sent is None:
+                assert rcvd_then_sent is None
+                msg = f"received {rcvd}; no close frame sent"
+            else:
+                assert rcvd_then_sent is not None
+                if rcvd_then_sent:
+                    msg = f"received {rcvd}; then sent {sent}"
+                else:
+                    msg = f"sent {sent}; then received {rcvd}"
```
cjerdonek, line 109: My main comment / question at this point is about the decision to construct the message in the `__init__()` method as opposed to doing it inside an `__str__()`. Also, it seems like it would be worth adding a `__repr__()` reflecting the underlying attributes.

aaugustin: I'm making the change for `__str__`. To be honest I was just continuing with the style I used until now.

It looks like we're getting a good `__repr__` for free:

```pycon
>>> from websockets.exceptions import ConnectionClosed
>>> from websockets.frames import Close
>>> ConnectionClosed(Close(1000, ""), Close(1000, ""), True)
ConnectionClosed(Close(code=1000, reason=''), Close(code=1000, reason=''), True)
```

cjerdonek: Regarding the `__repr__`, is that just using the initial `args`? If so, will it reflect the information correctly if one of the attributes is passed as a keyword argument, or if the attribute is modified after the fact?

aaugustin: Indeed `__repr__` appears to be set at `__init__` time. I don't see the use case for modifying an exception or even for instantiating it outside of websockets so I don't think it's a concern.

PR #1032, opened 2021-08-16
enforce ascii encoding for headers

lewoudar:
Even if in the code, it is said that headers contain only ascii characters, the default encoding used is `utf-8`, so I changed it to be `ascii`

### DIFF
--- a/src/websockets/datastructures.py
+++ b/src/websockets/datastructures.py
@@ -94,7 +94,7 @@ class Headers(MutableMapping[str, str]):
 
     def serialize(self) -> bytes:
         # Headers only contain ASCII characters.
-        return str(self).encode()
+        return str(self).encode('ascii')
 
     # Collection methods
 


aaugustin:
Hello! Can you clarify what issue you're trying to solve? This will help me understand if this is the best way to fix it.

For example, if the issue is "I think websockets should crash rather than send a invalid HTTP header if a user sets a non-ASCII header", I'd like to consider crashing when the user sets the header rather than when websockets sends the HTTP request or response. I'd also consider encoding the header properly rather than crashing. To make the correct decision there, I need to understand the use case.

My question also comes from your choice to fix only one instance of this pattern when there's at least half a dozen other instances in the code base, even if you ignore the `legacy` package.

lewoudar:
Hi Augustin,
My concern is only for the HTTP headers, not for all the `encode` methods used in the project. I just saw it when reading your code.
It is not really an issue if you decide to support `utf-8`, but your comment before the code I modified said that headers are `ascii` encoded, but when you use `str.encode` in python without specifying an encoding, it is `utf-8` by default. Readers can therefore be confused when reading the code. This is why I proposed this change.
But actually, just the comment can be changed :)

aaugustin:
Yes, I'm aware that UTF-8 is the default in Python 3. I don't usually specify ASCII in this case because that keeps the code short. It's clear to me, but I see how it could be confusing to someone else.

websockets itself only generates ASCII inputs to this function; however users may (incorrectly) generate non-ASCII inputs. I'm slightly worried about making a change that would create a hard-to-debug crash for these users on code that previously didn't crash (even though it behaved incorrectly).

Also `encode()` without an argument is marginally faster:

```
$ python -m timeit '''"Basic asntoehbuasntoehusanheubsaeo".encode()'''
5000000 loops, best of 5: 67.2 nsec per loop

$ python -m timeit '''"Basic asntoehbuasntoehusanheubsaeo".encode("ascii")'''
5000000 loops, best of 5: 83.7 nsec per loop
```

Since the concern is primarily about maintainability, I'm leaning towards clarifying the comment.

lewoudar:
Thanks for the reply and also to consider changing the comment

Issue #1038, opened 2021-08-31
TypeError: analyze() takes 1 positional argument but 2 were given


pourfard:
https://github.com/aaugustin/websockets/blob/1b0b2de59f13b79062a888a8d17ae5e9e5a60ef9/src/websockets/legacy/server.py#L305

Removing path as the second parameter resolves the issue.

aaugustin:
I guess you have:

```python
async def analyze(websocket):
    ...
```

somewhere and it should really be:

```python
async def analyze(websocket, path):
    ...
```

(In a future version of the API, I'm thinking of switching to the former, but for now the API is the latter.)

pourfard:
Thanks.

Yes, I changed it and the there is no problem.

PR #1045, opened 2021-09-09
Remove path argument from connection handlers.

aaugustin:
Ensure backwards-compatibility.

Ref #1038.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -63,7 +63,7 @@ And here's an echo server:
     import asyncio
     from websockets import serve
 
-    async def echo(websocket, path):
+    async def echo(websocket):
         async for message in websocket:
             await websocket.send(message)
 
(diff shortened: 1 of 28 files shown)


chrysn:
Just as a side note for future compatible changes: The inspection based remove_path_argument does not work flawlessly, and fail for handlers that were created through `functools.partial`. (To be fair, I'm not sure such a change can be done in Python in an API preserving way; given the tools this probably even works most of the time, just not all the times).

aaugustin:
Uh oh. That's pretty bad because the docs [recommend functools.partial](https://websockets.readthedocs.io/en/stable/howto/faq.html?highlight=functools#how-can-i-pass-additional-arguments-to-the-connection-handler). Thanks for the heads-up.

aaugustin:
@chrysn Fixed in d2808dc. I'm planning to make a release in the next few days (worst case by the end of the month).

PR #1126, opened 2022-02-01
Don't rely on `git` being on the path

csobrinho:
Allow version check to ignore `git` if the command is not available on the path.

### DIFF
--- a/src/websockets/version.py
+++ b/src/websockets/version.py
@@ -42,7 +42,7 @@ if not released:  # pragma: no cover
                 check=True,
                 text=True,
             ).stdout.strip()
-        except subprocess.CalledProcessError:
+        except (subprocess.CalledProcessError, FileNotFoundError):
             pass
         else:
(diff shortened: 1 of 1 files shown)


aaugustin:
I assumed you would have git around if you're working on a git clone — the only way to get a version of the code where `released = False`. Probably I missed a use case. How did you end up in this situation?

Depending on the situation, maybe the exception catching should be even broader.

csobrinho:
Hi @aaugustin we import the library into our repository but then when the software uses it we don't have access to git or if we have it, it won't be on the path so we would need the full path.

aaugustin:
I see - you're vendoring an unreleased version. That's a legitimate use case if you want to use a new feature before it's released.

csobrinho:
Great, let me know if you need anything extra for this PR. Thanks!

aaugustin:
The need is clear. I'm at best mildly satisfied of the current version management. I'll see if I just merge your proposal or if I reconsider the approach.

PR #1133, opened 2022-02-26
Update documentation

evanraalte:
Fix misleading function definition for `play` function in intro/tutorial2


aaugustin:
Thanks! Indeed, I didn't keep that section of the docs in sync with the code at some point when I was writing the tutorial.

PR #1134, opened 2022-02-28
fix logging error when send a memoryview

wwqgtxx:
ref https://bugs.python.org/issue15945

### DIFF
--- a/src/websockets/frames.py
+++ b/src/websockets/frames.py
@@ -139,7 +139,7 @@ class Frame:
             # Encode just what we need, plus two dummy bytes to elide later.
             binary = self.data
             if len(binary) > 25:
-                binary = binary[:16] + b"\x00\x00" + binary[-8:]
+                binary = b''.join([binary[:16], b"\x00\x00", binary[-8:]])
             data = " ".join(f"{byte:02x}" for byte in binary)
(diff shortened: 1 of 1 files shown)


aaugustin:
Thanks for the pull request.

 I would like to add a test for this (because it would be too easy to create a regression otherwise).

wwqgtxx:
Maybe we can simply copy and modify this test code
https://github.com/aaugustin/websockets/blob/498cc8c061e53f0001cb2e3ade22ee8ce5ff11a1/tests/test_frames.py#L273-L278
to add a new test function like this
```python
 def test_binary_memoryview_truncated(self): 
     self.assertEqual( 
         str(Frame(OP_BINARY, memoryview(bytes(range(256))))), 
         "BINARY 00 01 02 03 04 05 06 07 08 09 0a 0b 0c 0d 0e 0f ..." 
         " f8 f9 fa fb fc fd fe ff [256 bytes]", 
     ) 
```
But maybe we need more test code to cover other places in the code where `memoryview` or even `bytearray` might be passed.

aaugustin:
Yes it's easy :-) I'll take care of it.

aaugustin:
I rebased, added tests in 778879e, and merged. Thank you for your contribution!

PR #1145, opened 2022-03-14
README: Add path parameter to echo handler

Sh4pe:
Without this fix, the example from README does not work.


aaugustin:
You are reading the docs for websockets ≥ 10.1 while using websockets < 10.1.

Sh4pe:
Oh, ok, my bad. Sorry about that!

Von meinem iPhone gesendet

PR #1147, opened 2022-03-15
Add ':' charter in allowed symbols for _token_re

livestalker:
':' is allowed symbol for subprotocol name.
For example: "ogmios.v1:compact"

### DIFF
--- a/src/websockets/headers.py
+++ b/src/websockets/headers.py
@@ -93,7 +93,7 @@ def parse_OWS(header: str, pos: int) -> int:
     return match.end()
 
 
-_token_re = re.compile(r"[-!#$%&\'*+.^_`|~0-9a-zA-Z]+")
+_token_re = re.compile(r"[-!#$%&\'*+.^_`|~:0-9a-zA-Z]+")
 
 
 def parse_token(header: str, pos: int, header_name: str) -> Tuple[str, int]:


aaugustin:
I disagree with your statement that `:` is allowed in subprotocols.

Per https://datatracker.ietf.org/doc/html/rfc6455#section-4.1:

> The request MAY include a header field with the name
        |Sec-WebSocket-Protocol|.  If present, this value indicates one
        or more comma-separated subprotocol the client wishes to speak,
        ordered by preference.  The elements that comprise this value
        MUST be non-empty strings with characters in the range U+0021 to
        U+007E not including separator characters as defined in
        [[RFC2616](https://datatracker.ietf.org/doc/html/rfc2616)] and MUST all be unique strings.  The ABNF for the
        value of this header field is 1#token, where the definitions of
        constructs and rules are as given in [[RFC2616](https://datatracker.ietf.org/doc/html/rfc2616)].

Per https://datatracker.ietf.org/doc/html/rfc2616#section-2.2:

```
       token          = 1*<any CHAR except CTLs or separators>
       separators     = "(" | ")" | "<" | ">" | "@"
                      | "," | ";" | ":" | "\" | <">
                      | "/" | "[" | "]" | "?" | "="
                      | "{" | "}" | SP | HT
```

Sec-WebSocket-Protocol contains tokens. `:` is explicitly listed as a separator which isn't allowed in tokens.

Furthermore, all subprotocols registered in the [subprotocol registry](https://www.iana.org/assignments/websocket/websocket.xml) respect the rules.

Besides, this is dangerously close to https://websockets.readthedocs.io/en/stable/project/contributing.html#cryptocurrency-users

PR #1155, opened 2022-04-01
Add a new line about client.connect() and asyncio.loop.create_connection() accept the same arguments

redphx:
It wasn't very clear in [documentation](https://websockets.readthedocs.io/en/stable/reference/client.html), and I didn't know about this until I see the same line in this [FAQ section](https://websockets.readthedocs.io/en/stable/howto/faq.html#how-do-i-disable-tls-ssl-certificate-verification). I didn't know I could pass `sock` parameter to `client.connect()` because it wasn't mentioned.

### DIFF
--- a/src/websockets/legacy/client.py
+++ b/src/websockets/legacy/client.py
@@ -344,6 +344,8 @@ class Connect:
     """
     Connect to the WebSocket server at ``uri``.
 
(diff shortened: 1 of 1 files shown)


aaugustin:
It is already mentioned a few lines below: https://github.com/aaugustin/websockets/pull/1155/files#diff-4cdef9c9e58dd4489278768e7e5e71a928bc04f0b97817419c5904bcc9e58817R405-R406

Issue #1175, opened 2022-05-15
Python 3.11 test failures


mgorny:
When running the test suite using Python 3.11.0b1, I get the following test failures/errors:

```pytb
$ tox -e py311
GLOB sdist-make: /tmp/websockets/setup.py
py311 create: /tmp/websockets/.tox/py311
py311 inst: /tmp/websockets/.tox/.tmp/package/1/websockets-10.4.dev13+gc4a4b6f.zip
py311 installed: websockets @ file:///tmp/websockets/.tox/.tmp/package/1/websockets-10.4.dev13%2Bgc4a4b6f.zip
py311 run-test-pre: PYTHONHASHSEED='417100899'
py311 run-test: commands[0] | python -W default -m unittest
....................................................................................................................................................................F.................................................................EE................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................
======================================================================
ERROR: test_client (tests.legacy.test_client_server.YieldFromTests.test_client)
----------------------------------------------------------------------
Traceback (most recent call last):
  File "/tmp/websockets/tests/legacy/test_client_server.py", line 128, in _decorate
    return func(self, *_args, **_kwargs)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/tmp/websockets/tests/legacy/test_client_server.py", line 1308, in test_client
    @asyncio.coroutine
     ^^^^^^^^^^^^^^^^^
AttributeError: module 'asyncio' has no attribute 'coroutine'

======================================================================
ERROR: test_server (tests.legacy.test_client_server.YieldFromTests.test_server)
----------------------------------------------------------------------
Traceback (most recent call last):
  File "/tmp/websockets/tests/legacy/test_client_server.py", line 1322, in test_server
    @asyncio.coroutine
     ^^^^^^^^^^^^^^^^^
AttributeError: module 'asyncio' has no attribute 'coroutine'

======================================================================
FAIL: test_explicit_socket (tests.legacy.test_client_server.SecureClientServerTests.test_explicit_socket)
----------------------------------------------------------------------
Traceback (most recent call last):
  File "/tmp/websockets/tests/legacy/test_client_server.py", line 128, in _decorate
    return func(self, *_args, **_kwargs)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/tmp/websockets/tests/legacy/test_client_server.py", line 452, in test_explicit_socket
[...]

PR #1176, opened 2022-05-15
Fix Python 3.11 test failures

mgorny:
Fix the test failures reported in #1175.


hroncok:
Thank you.

aaugustin:
Thank your for fixing the tests! See #1211 for the rest of adding Python 3.11.

Issue #940, opened 2021-05-04
Release 9.0 breaks mypy


vertti:
Running a project that uses websockets 9.0 against mypy, throws:

```
error: Module has no attribute "connect"
```
for 
```python
import websockets
await websockects.connect(uri)
```

and
```
Module 'websockets' has no attribute 'connect'
```
for 
```python
from websockets import connect
```

Version 8.1 of websockets has no such issue. Using the latest mypy 0.812.

apljungquist:
I expect this is caused by the new lazy import system. Another consequence of this is that trying to jump to definitions no longer works (one always lands in the package dunder init file) which is unfortunate since the ability to quickly inspect third party code is one of my favourite features of python.

aaugustin:
Oh, the joy.

gmjosack:
The lazy importer also breaks pylance, similar to the error for mypy, as well as pyinstaller as it fails to include the correct modules in the distribution. 

as a work around I'm currently using 

```py
from websockets.legacy.client import connect as ws_connect
```

which resolved both issues.

aaugustin:
This is the pragmatic workaround at this point. I hate that it exposes the `.legacy` package before the new package is ready, but I'm not coming up with something better.

https://stackoverflow.com/questions/60739889/bypass-mypys-module-has-no-attribute-on-dynamic-attribute-setting has a good discussion of the options and they're all bad :-(

aaugustin:
Could you report whether https://github.com/aaugustin/websockets/commit/a7b240064b579f0c1a5b14d5404646bf289092f3 works?

Supported uses should include:

```python
import websockets.client
await websockects.client.connect(uri)
```

and

```python
from websockets.client import connect
await connect(uri)
```

(This will have to be a bugfix release on top on 9.0.1, but I never did maintenance branches in websockets, so this isn't a request.)

ltworf:
You could have at least put a changelog file to say that the api was changing.

Also semver would be nice.

ltworf:
In any case, changing the import seems to work but doesn't satisfy mypy.

`slackclient/client.py:63: error: Name 'websockets.client.WebSocketClientProtocol' is not defined`

ltworf:
Well imports are code.

Specifically now I have the problem that distributions use one version, pip installs another and I don't know if I can make code compatible with both.

Also mypy fails even after fixing the imports so that doesn't help so much.

aaugustin:
@ltworf -

[...]

PR #1191, opened 2022-06-26
Add explanation of imports to API reference

0xdade:
Yesterday, while trying to type check my code using the `websockets` library, I was getting quite frustrated. All of the example code seemed to show `import websockets` and then accessing everything directly off the package, which doesn't support type checking or code completion because of the lazy import system. I nearly opened an issue about the matter to see what I was doing wrong, when I then went to double check that there wasn't an existing issue about it before filing. Lo and behold, I found https://github.com/aaugustin/websockets/issues/940#issuecomment-874012438 which gave me what I needed to start type checking my code that uses `websockets`.

I couldn't find this directly provided as an example or in any of the example code I saw in the docs, but while digging around did see this reference to "use the real import path" at the bottom of the API reference index, without really elaborating on what that is. I consider myself a reasonably experienced python dev and was surprised/confused at first to see `import websockets` not exposing any of the modules to my IDE but `from websockets import ` exposing all available modules (and thereby giving my autocomplete and letting mypy type check correctly).

So rather than just raise another issue complaining that I couldn't figure it out, I thought I'd add in a couple more lines to the docs to explain what "use the real import paths" actually looks like in code.

I also considered perhaps adding a note to the [types](https://websockets.readthedocs.io/en/stable/reference/types.html) page of the docs, but felt that wasn't really the correct place since that's a proper API reference. I also thought about maybe a new docs page in the [how-to guides](https://websockets.readthedocs.io/en/stable/howto/index.html) about how to use the websockets library in a way that your IDE and type checkers know what's going on, but thought I'd start with this simple change first.

### DIFF
--- a/docs/reference/index.rst
+++ b/docs/reference/index.rst
@@ -61,6 +61,8 @@ guarantees of behavior or backwards-compatibility for private APIs.
 For convenience, many public APIs can be imported from the ``websockets``
 package. However, this feature is incompatible with static code analysis. It
 breaks autocompletion in an IDE or type checking with mypy_. If you're using
-such tools, use the real import paths.
+such tools, use the real import paths. To do this, you must import modules via
+``from websockets import <module>`` or to access a specific exported value
+``from websockets.<module> import <exported_value>``.
 
 .. _mypy: https://github.com/python/mypy


povilasb:
Thank you 🙌 
It surely helped me.

aaugustin:
Addressed in 206d7ef. I chose to link to existing documentation rather than duplicate information.

aaugustin:
Thank you for the suggestion!

PR #1205, opened 2022-08-10
Default values for Optional fields in WebSocketURI

treylav:
Hello!

Due to the nature of dataclasses in Python, for a field to be truly optional, it must be provided with a default value, not only `Optional` annotation. Otherwise, when creating `WebSocketURI` сlass instance without `login` and `password`, we will get an error like:
```
TypeError: WebSocketURI.init() missing 2 required positional arguments: 'username' and 'password'
```
Judging by the code, this was an accidental oversight, because checking for `None` value of the `name` and `password` already exists and the rest of the project code has default values for dataclasses.

Thanks in advance.

### DIFF
--- a/src/websockets/uri.py
+++ b/src/websockets/uri.py
@@ -33,8 +33,8 @@ class WebSocketURI:
     port: int
     path: str
     query: str
-    username: Optional[str]
-    password: Optional[str]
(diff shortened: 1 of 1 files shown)


aaugustin:
Thank you!

PR #1211, opened 2022-08-17
Add Python 3.11.

aaugustin:


.github/workflows/tests.yml
```diff
@@ -43,6 +43,7 @@ jobs:
           - "3.8"
           - "3.9"
           - "3.10"
+          - "3.11"
```
Kludex, line 46: Or... Are you waiting for 3.11?
```suggestion
          - "3.11-dev"
```

aaugustin: Yes, I usually merge such PRs when the target version is available on GitHub Actions. I confirmed locally that 3.11 works. The risk of introducing a change that works on 3.10 but not 3.11 are vanishingly small.

aaugustin:
I merged most of this to `main` in separate commits. The only remaining task is to enable 3.11 in CI, which I don't want to do now due to https://github.com/actions/setup-python/pull/414.

PR #1220, opened 2022-08-23
Configure ClusterFuzzLite.

aaugustin:
- Copy OSS-Fuzz configuration.
- Add ClusterFuzzLite GitHub Action.


aaugustin:
Note - remove the [TEMPORARY] commit before merging.

aaugustin:
Given that the parsing code is almost never changed, I don't see much value in burning lots of CPU cycles on all pull requests, so I'm probably not going to merge this. https://github.com/google/oss-fuzz/pull/8339 testing the main branch would be enough.

Issue #1226, opened 2022-09-11
Multiple headers with same name


iudeen:
Is setting multiple headers with same name not allowed when using the legacy server?

Currently, when we set multiple headers with same name, an exception is raised.

Is there a way to overcome this?

aaugustin:
Can you provide the full stack trace please?

iudeen:
```
Traceback (most recent call last):
  File "D:\ProgramData\Anaconda3\envs\uvicorn\lib\site-packages\websockets\legacy\server.py", line 165, in handler
    await self.handshake(
  File "D:\ProgramData\Anaconda3\envs\uvicorn\lib\site-packages\websockets\legacy\server.py", line 643, in handshake
    response_headers.setdefault("Server", self.server_header)
  File "D:\ProgramData\Anaconda3\envs\uvicorn\lib\_collections_abc.py", line 1007, in setdefault
    return self[key]
  File "D:\ProgramData\Anaconda3\envs\uvicorn\lib\site-packages\websockets\datastructures.py", line 119, in __getitem__
    raise MultipleValuesError(key)
```

aaugustin:
It looks like `Headers.setdefault` doesn't work when a key already has multiples values. That would be a bug.

(Out of curiosity, what is the use case for setting multiple `Server` headers? I've never seen that.)

iudeen:
Thanks for the reply.

> It looks like `Headers.setdefault` doesn't work when a key already has multiples values. That would be a bug.

Oh, if that's the case, can we fix this with a PR? I can try to help.

> (Out of curiosity, what is the use case for setting multiple `Server` headers? I've never seen that.)

Well, this emerged while working on a PR in Uvicorn. Uvicorn allows users to specify their own headers. This is to cover a scenario where they might add `Server` header.

I personally don't think of, or have seen, any real any benefit in setting multiple `Server` header.

aaugustin:
As mentioned in the merge request, the current behavior of `Headers.setdefault` is correct. We cannot change it because it returns the current value when present and we don't want to do this when there are multiple values.

Given that there's no use case for multiple `Server` or `User-Agent` headers, I don't see any reason for spending more time on this.

PR #1227, opened 2022-09-11
Allow setting multiple server headers in legacy server implementation 

iudeen:
Fixes: https://github.com/aaugustin/websockets/issues/1226

### DIFF
--- a/src/websockets/legacy/server.py
+++ b/src/websockets/legacy/server.py
@@ -640,7 +640,7 @@ class WebSocketServerProtocol(WebSocketCommonProtocol):
 
         response_headers.setdefault("Date", email.utils.formatdate(usegmt=True))
         if self.server_header is not None:
-            response_headers.update("Server", self.server_header)
+            response_headers.update({"Server": self.server_header})
 
         self.write_http_response(http.HTTPStatus.SWITCHING_PROTOCOLS, response_headers)
 
--- a/src/websockets/legacy/server.py
+++ b/src/websockets/legacy/server.py
@@ -640,7 +640,7 @@ class WebSocketServerProtocol(WebSocketCommonProtocol):
 
         response_headers.setdefault("Date", email.utils.formatdate(usegmt=True))
         if self.server_header is not None:
-            response_headers.update({"Server": self.server_header})
+            response_headers["Server"] = self.server_header
 
         self.write_http_response(http.HTTPStatus.SWITCHING_PROTOCOLS, response_headers)
 
--- a/src/websockets/legacy/server.py
+++ b/src/websockets/legacy/server.py
@@ -633,14 +633,16 @@ class WebSocketServerProtocol(WebSocketCommonProtocol):
         if protocol_header is not None:
             response_headers["Sec-WebSocket-Protocol"] = protocol_header
 
-        if callable(extra_headers):
-            extra_headers = extra_headers(path, self.request_headers)
-        if extra_headers is not None:
-            response_headers.update(extra_headers)
+        if self.server_header is not None:
+            response_headers.setdefault("Server", self.server_header)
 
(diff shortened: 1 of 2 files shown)


src/websockets/legacy/server.py
```diff
@@ -640,7 +640,7 @@ async def handshake(
 
         response_headers.setdefault("Date", email.utils.formatdate(usegmt=True))
         if self.server_header is not None:
-            response_headers.setdefault("Server", self.server_header)
+            response_headers["Server"] = self.server_header
```
aaugustin, line 643: You cannot do this: it's a backwards incompatible change.

aaugustin (changes requested):

src/websockets/legacy/server.py
iudeen: Oh, so only way is to use `setdefault`?

I assume `response_headers.update({"Server": self.server_header})` will cause backward incompatibility..

aaugustin: I'm leaning towards fixing the `Headers.setdefault` method.

aaugustin: But I'd have to look at the implementation of the parent class to check that we aren't headed down a rabbit hole if we do that...

iudeen: I am not sure if `Headers` overrides `setdefault`. It's a method of MutableMapping I suppose? (I might be wrong)

iudeen: Just a question out of curiosity, if it's okay.. 

What's the backwards incompatibility here? Aren't we setting value to the Mapping object in both cases?

aaugustin: The backwards-incompatibility in your proposal is that you are overriding any previously set value. Using `setdefault` here was an escape hatch to support customizing the `Server` header before we formalized an API to do that (in the change we made together recently).

iudeen: Thanks for the reply!

I found while testing that when we set headers directly (like I did in this PR), it doesn't replace, rather append another header to it. That is, if we have `Server` header already and if user sets another, we will end up having 2 of those.

iudeen: I rearranged and made setting of server header before setting extra headers. This handles the scenario well, without making any change. 

Is this a better solution?

iudeen: >If you set this header (by overriding private APIs, probably...) websockets doesn't override it.

 Another observation:

 If I do this 
```python
        if self.server_header is not None:
            response_headers.setdefault("Server", self.server_header)
            response_headers["Server"] = "Potato"
```
we get
```
Headers([('Upgrade', 'websocket'), ('Connection', 'Upgrade'), ('Sec-WebSocket-Accept', '9i+pz2BTKb6LxQmkX4v7j7g71ps='), ('Server', 'Python/3.10 websockets/10.4.dev47+g99c8440.dirty'), ('Server', 'Potato'), ('Date', 'Mon, 12 Sep 2022 16:14:46 GMT')])
```
Which means doing ` response_headers["Server"] = "Potato"` adds another header, and not override right?

aaugustin: > I rearranged and made setting of server header before setting extra headers. This handles the scenario well, without making any change.

> Is this a better solution?

I don't think so: you're breaking the code of those who set `Server` via `extra_headers`.

aaugustin: Changing `Headers.setdefault` isn't possible because this methods returns the value and we don't want to do this when there are multiple values.

iudeen: > Changing `Headers.setdefault` isn't possible because this methods returns the value and we don't want to do this when there are multiple values.

Agreed. It'd affect the normal behavior.

aaugustin:
See #1226.

PR #1245, opened 2022-10-16
Add sockets & threads implementation

aaugustin:


aaugustin:
I reached 100% coverage. I think the implementation is in a reasonable shape. There are still a few things to clean up and documentation to add.

PR #1256, opened 2022-11-01
pylint fixes

marksmayo:
Simplified some branching structures, reordered imports


aaugustin:
Thanks for the proposal but I disagree:

* I don't want to clutter the repo with editor-specific config files.
* The proposed change to import order are net negative.
* There is no value in changing `if/elif/else` style (same feeling expressed in you last similar PR: https://github.com/webpy/webpy/pull/740)

I recommend that you start a discussion investing effort in cosmetic changes in a repository where you don't have a track record of contributing.

PR #1267, opened 2022-11-23
Add CIFuzz GitHub action

DavidKorczynski:
Add [CIFuzz](https://google.github.io/oss-fuzz/getting-started/continuous-integration/) workflow action to have fuzzers build and run on each PR.

This is a service offered by OSS-Fuzz where websockets was recently integrated. CIFuzz can help detect regressions and catch fuzzing build issues early, and has a variety of features (see the URL above). In the current PR the fuzzers gets build on a pull request and will run for 300 seconds.

### DIFF
--- a/.github/workflows/cifuzz.yml
+++ b/.github/workflows/cifuzz.yml
@@ -1,0 +1,26 @@
+name: CIFuzz
+on: [pull_request]
+jobs:
+  Fuzzing:
+    runs-on: ubuntu-latest
+    steps:
+    - name: Build Fuzzers
+      id: build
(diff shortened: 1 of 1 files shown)


aaugustin:
Hello David, thank you for submitting this pull request.

When I submitted websockets to OSS Fuzz, I considered running ClusterFuzzLite. I decided against it because the code exercised by OSS Fuzz — the parsers — is extremely stable at this point.

While it's valuable to have OSS Fuzz explore it in depth, I don't see much value in burning CPU on every pull request when code covered by the fuzzing targets almost never changes. Also, I'm trying to keep CI as simple and short as possible.

Am I missing a benefit other than "discovering issues before merging to main, rather than after?"

PR #1274, opened 2022-12-10
Refactor create_protocol to represent the real type

ooliver1:
`[Any]` means one argument of `Any` type, `...` means "ignore". This is relevant since `WebSocketServerProtocol` and similar in fact have multiple arguments. Ideally a `Protocol` with a `__call__` method included should be used, but that's up for discussion since that would add maintainability issues as the signature would be duplicated in multiple places.

**Or**, `Type[WebSocketServerProtocol]` *may* work too.


ooliver1:
Been so long, I forgor about this PR

aaugustin:
Sorry for the delay on my end. I think it's interesting and want to review it. Just had other personal priorities in the recent months.

aaugustin:
Looks good. I made a few more changes in f42fd7ba. Thank you.

PR #1279, opened 2022-12-28
Added Custom Ping Payload on Client Connect to keep Connection alive

anil-kamani:
When someone wants to send a Custom Ping message on every ping interval, they can directly add it from ping_payload.


aaugustin:
What is the use case?

I'm asking because `connect()` has many parameters already and I don't want to add more without a strong reason. Adding a parameter creates overhead for everyone reading the documentation.

anil-kamani:
Use case: now day Many WebSocket Servers are Accepting only their Custom Ping/Heartbeat Message to Keep the Connection Alive.  i am working on many Projects which require me to send Custom Ping Payload instead of Random Binary Bytes.

aaugustin:
These servers require the heartbeat to be sent in a message of type "text", not of type "ping".

The example you're giving says _"Heartbeat request (text)"_ <-- text, not ping.

The change you are proposing here doesn't meet this requirement.

You should do something like this instead:

```python
def heartbeat(websocket):
    while True:
        await asyncio.sleep(25)
        await websocket.send("ping")

websocket = connect(...)
asyncio.create_task(heartbeat(websocket))
```

----

> and if you see the WebSocket-client library of python it also supports the Custom Ping Payload.

websockets has the same level of support for custom ping payloads as websocket-client. It offers the same API.

----

Perhaps websockets could grow a mechanism for configuring arbitrary keepalive mechanisms. But it's a broader change than this one.

anil-kamani:
Okey Sir, No Problem will Contribute to the Library if I have anything. Thank you.

ecoue:
It would be advantageous to be able to customize the application data sent in a ping frame. Although external clients/servers usually respond to a ping frame, they often do not do so in accordance with RFC 6455. In particular, the application data is regularly omitted, which is why the keepalive ultimately fails (see below). If it were possible to specify the transmission of empty application data in ping frames, which RFC 6455 explicitly permits (5.5.2.), the subsequent pong without application data would also comply with the specification (5.5.3.) and should therefore keep the connection alive.

```
DEBUG [2024-05-06 23:37:16] websockets.client - = connection is OPEN
DEBUG [2024-05-06 23:37:31] websockets.client - % sending keepalive ping
DEBUG [2024-05-06 23:37:31] websockets.client - > PING ea 97 7b 97 [binary, 4 bytes]
DEBUG [2024-05-06 23:37:31] websockets.client - < PONG '' [0 bytes]
DEBUG [2024-05-06 23:37:41] websockets.client - ! timed out waiting for keepalive pong
DEBUG [2024-05-06 23:37:41] websockets.client - ! failing connection with code 1011
DEBUG [2024-05-06 23:37:41] websockets.client - = connection is CLOSING
DEBUG [2024-05-06 23:37:41] websockets.client - > CLOSE 1011 (internal error) keepalive ping timeout [24 bytes]
DEBUG [2024-05-06 23:37:41] websockets.client - = connection is CLOSED
```

PR #1302, opened 2023-02-24
Set server_hostname automatically when needed

shafemtol:
`create_connection` requires either `host` or `server_hostname` if `ssl` is truthy. If `sock` is given, `host` gets unset, so provide `server_hostname` with the URI host if needed and not already provided by the caller.


aaugustin:
Thank you. I added tests & docs for this.

PR #1307, opened 2023-03-07
Switch to using asyncio.timeout instead of asyncio.wait_for

bdraco:
`asyncio.wait_for` creates a task whereas `asyncio.timeout` avoids doing this which decreases latency.

Fallback to using `async_timeout` when the python version is too old (<3.11)

`asyncio.timeout` will become the underlying implementation for `async.wait_for` in cpython 3.12
https://github.com/python/cpython/pull/98518

### DIFF
--- a/src/websockets/legacy/compatibility.py
+++ b/src/websockets/legacy/compatibility.py
@@ -5,6 +5,9 @@ import sys
 from typing import Any, Dict
 
 
+__all__ = ["asyncio_timeout", "loop_if_py_lt_38"]
+
+
 if sys.version_info[:2] >= (3, 8):
 
     def loop_if_py_lt_38(loop: asyncio.AbstractEventLoop) -> Dict[str, Any]:
@@ -22,3 +25,9 @@ else:
 
         """
         return {"loop": loop}
+
+
+if sys.version_info[:2] >= (3, 11):
+    from asyncio import timeout as asyncio_timeout  # noqa: F401
+else:
+    from .async_timeout import timeout as asyncio_timeout  # noqa: F401
--- a/tests/legacy/test_protocol.py
+++ b/tests/legacy/test_protocol.py
@@ -1579,7 +1579,8 @@ class ServerTests(CommonTests, AsyncioTestCase):
             self.receive_frame(self.close_frame)
             self.run_loop_once()
             self.loop.run_until_complete(self.protocol.close(reason="close"))
-        self.assertConnectionClosed(1000, "close")
+        # Due to a bug in coverage, this is erroneously reported as not covered.
+        self.assertConnectionClosed(1000, "close")  # pragma: no cover
 
     def test_local_close_connection_lost_timeout_after_close(self):
         self.protocol.close_timeout = 10 * MS
@@ -1596,7 +1597,8 @@ class ServerTests(CommonTests, AsyncioTestCase):
             self.receive_frame(self.close_frame)
             self.run_loop_once()
             self.loop.run_until_complete(self.protocol.close(reason="close"))
-        self.assertConnectionClosed(1000, "close")
+        # Due to a bug in coverage, this is erroneously reported as not covered.
+        self.assertConnectionClosed(1000, "close")  # pragma: no cover
 
 
 class ClientTests(CommonTests, AsyncioTestCase):
@@ -1614,7 +1616,8 @@ class ClientTests(CommonTests, AsyncioTestCase):
         # Check the timing within -1/+9ms for robustness.
         with self.assertCompletesWithin(19 * MS, 29 * MS):
             self.loop.run_until_complete(self.protocol.close(reason="close"))
-        self.assertConnectionClosed(1006, "")
+        # Due to a bug in coverage, this is erroneously reported as not covered.
+        self.assertConnectionClosed(1006, "")  # pragma: no cover
 
     def test_local_close_receive_close_frame_timeout(self):
         self.protocol.close_timeout = 10 * MS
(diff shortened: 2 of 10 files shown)


aaugustin:
Thank you for the suggestion! I think it's a good idea. Let's do it.

Until now, websockets has no dependencies. This removes an entire category of concerns like managing versions that, as a single maintainer, I don't want to deal with. I understand that this may not look like a big benefit but it is valuable to me.

In order to keep this benefit, I am tempted to do one of the following:

1. create a small compatibility layer to use asyncio.wait_for on Python < 3.11 and asyncio.timeout on Python ≥ 3.12
2. vendor CPython's or  async-timeout's implementation — they're probably identical if async-timeout was merged into CPython.

Option 1 could be as simple as:

```python
if sys.version_info[:2] >= (3, 12):

    async def asyncio_wait_for(coro, timeout):
        async with asyncio.timeout(timeout):
            await coro

else:

    asyncio_wait_for = asyncio.wait_for
```

In all cases, version compatibility code should go in [websockets.legacy.compatibility](https://github.com/aaugustin/websockets/blob/main/src/websockets/legacy/compatibility.py) rather than a new module.

aaugustin:
I'm aware that my option 1 doesn't yield performance benefits on Python < 3.12 while your version does. IMO those who care about performance optimizations available in Python 3.12 should upgrade to Python 3.12. It isn't my job to fake that for them.

bdraco:
I'll work on option 2 since the races in `async.wait_for` and the issue tickets they generated for `cpython` were part of the motivation for cpython switching out the implementation.  It think it would be best if everything behaved the same regardless of the python version.

ex:
https://github.com/python/cpython/issues/86296
https://github.com/python/cpython/issues/81839
https://github.com/python/cpython/issues/96764

bdraco:
Its passing locally now. Hopefully it should be good to go now

tests/legacy/test_protocol.py
```diff
@@ -929,6 +929,7 @@ def test_answer_ping_does_not_crash_if_connection_closed(self):
         self.receive_frame(Frame(True, OP_PING, b"test"))
         self.receive_eof()
 
+        self.loop.run_until_complete(asyncio.sleep(MS))
```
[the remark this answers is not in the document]
aaugustin: I'm not surprised that you're finding this sort of issue in the test suite.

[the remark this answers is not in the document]
aaugustin: It would be best to use `self.run_loop_once()`. It's faster than waiting 1ms.

Yes, it's just 1ms, but over hundreds of tests, delays add up!

src/websockets/legacy/async_timeout/__init__.py
aaugustin, line 1: I would like to put the code in `src/websockets/legacy/async_timeout.py`. We have no reason to nest it in under `async_timeout/__init__.py`.

src/websockets/legacy/compatibility.py
```diff
@@ -22,3 +22,12 @@ def loop_if_py_lt_38(loop: asyncio.AbstractEventLoop) -> Dict[str, Any]:
 
         """
         return {"loop": loop}
+
+
+if sys.version_info[:2] < (3, 11):
+    from .async_timeout import timeout as asyncio_timeout  # pragma: no cover
+else:  # pragma: no cover
```
aaugustin, line 29: I made changes in 399db68d so you can drop the `# pragma: no cover` on the two lines above.

src/websockets/legacy/compatibility.py
```diff
@@ -22,3 +22,12 @@ def loop_if_py_lt_38(loop: asyncio.AbstractEventLoop) -> Dict[str, Any]:
...
+    from asyncio import timeout as asyncio_timeout
+
+
+__all__ = ["asyncio_timeout", "loop_if_py_lt_38"]
```
aaugustin, line 33: This isn't needed because aren't doing `from websockets.legacy.compatibility import *` anywhere.

Did I miss a reason why it would be useful? Or did you just add it for consistency?

tox.ini
```diff
@@ -18,13 +18,13 @@ commands = python -W error::DeprecationWarning -W error::PendingDeprecationWarni
 commands =
     python -m coverage erase
     python -m coverage run --source {envsitepackagesdir}/websockets,tests -m unittest {posargs}
-    python -m coverage report --show-missing --fail-under=100
+    python -m coverage report --show-missing --omit src/websockets/legacy/async_timeout/__init__.py  --fail-under=100
```
aaugustin, line 21: I made changes in 399db68d to configure this in setup.cfg rather than in all coverage commands. It should be possible to remove this change.

aaugustin (commented):
Thank you again for the pull request. We're close to being able to merge it.

I left a few comments to polish it. Let me know if you'd like to do it or if I should finalize.

If you think this optimization deserves an entry in changelog.rst, feel free to add it.

aaugustin:
Also, It would be great if you could rebase and squash into a final commit to clean up the PR.

bdraco:
Late here.  I'll take a look in the morning if the Home Assistant beta isn't too busy.

src/websockets/legacy/compatibility.py
aaugustin: Ah, it avoids this warning from mypy:

src/websockets/legacy/protocol.py:56: error: Module "websockets.legacy.compatibility" does not explicitly export attribute "asyncio_timeout"  [attr-defined]

aaugustin:
I squashed your commits and I'm finalizing the changes here: #1325.

aaugustin:
I filed https://github.com/nedbat/coveragepy/issues/1598 for the coverage problem.

Thank you for your contribution!

bdraco:
I had this on my followup list and just returned to it and found you already got it over the finish line.  Many thanks 👍 🐬 🥇

aaugustin:
This caused https://github.com/encode/uvicorn/issues/1927.

bdraco:
Looks like you have already determined whats going on in https://github.com/encode/uvicorn/issues/1927#issuecomment-1656580406

Let me know if I can help in any way.

Cheers

PR #1320, opened 2023-04-01
Improve error handling in broadcast().

### LINKED ISSUE #1319: Insufficient error handling in broadcast
If broadcast fails to write the frame to a connection — for any reason — it should still write to all other connections.

The current implementation doesn't guarantee that because `broadcast` calls `self.write_frame_sync` which calls `self.transport.write` and there's no exception handling along the way. If `self.transport.write` fails for any reason, the exception will bubble.

Reported by @jlaine.

aaugustin:
Fix #1319.

### DIFF
--- a/src/websockets/legacy/protocol.py
+++ b/src/websockets/legacy/protocol.py
@@ -7,6 +7,7 @@ import logging
 import random
 import ssl
 import struct
+import sys
 import time
 import uuid
 import warnings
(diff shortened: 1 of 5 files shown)


src/websockets/legacy/protocol.py
```diff
@@ -1573,54 +1574,86 @@ def eof_received(self) -> None:
...
+    fragmentation is useful for sending large messages without buffering them in
+    memory, while :func:`broadcast` buffers one copy per connection as fast as
+    possible.
+
+    :func:`broadcast` skips connections that aren't open in order to avoid
+    errors on connections where the closing handshake is in progress.
+
+    :func:`broadcast` ignores failures to write the message. On Python 3.11 and
+    above, you may set ``raise_exceptions`` to :obj:`True` to raise all
+    exceptions in a :pep:`654` :exc:`ExceptionGroup` instead.
```
jlaine, line 1611: Maybe the wording can be refined to clarify the fact that `broadcast` will *always* continue writing to other websockets even if one fails. The `raise_exceptions` argument only controls whether the exceptions are logged as they occur or raised at the *end* of the run in an `ExceptionGroup`.

jlaine: The latest wordings looks great!

PR #1332, opened 2023-04-07
New asyncio implementation

aaugustin:


stalkerg:
Do you know if I can read anywhere for the purpose of this PR? Is it porting the current asyncio implementation over Sans IO?

aaugustin:
It is indeed aimed at providing an asyncio implementation on top of the Sans-I/O implementation.

Based on my experience of the threading implementation, ETA for completion is likely to be 1 to 4 years for now so don't hold your breath :-)

No significant user-visible improvements are expected. The benefits are mostly maintainer-visible.

PR #1349, opened 2023-04-29
Use selectors instead of select.poll in sync.WebSocket Server for multi-platform support

ceharris:
For a senior-level university network apps design course, my students were using the threaded version of `WebSocketServer` as part of their final project. While the project required their work to run successfully in containers on Linux, many of them use Windows for their development workstation. We discovered that `serve_forever` throws an error at startup on Windows due to the use of `select.poll` as the means to block while waiting for incoming socket connections. The documentation for `poll` indicates that it is not supported on all platforms, and apparently Windows is one such platform.

I patched `WebSocketServer` to instead use `selectors.DefaultSelector` which determines the best supported mechanism for I/O multiplexing on the runtime platform. The functionality is the same, but it works on a wider range of platforms.

After switching to `selectors.DefaultSelector`, I discovered that Windows also doesn't support I/O multiplexing using pipes or files -- only sockets. I removed the use of `os.pipe` for the shutdown mechanism. It was redundant anyway -- simply closing the listener socket in the `shutdown` method is sufficient to cause the selector to return. Subsequently, the call to `socket.accept` (on the closed listener socket) causes the loop in `serve_forever` to terminate as expected.

I tested the change on Windows, Mac OS X, and a Linux container, and it seems that it correctly supports all three platforms.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -61,6 +61,7 @@ source = [
 exclude_lines = [
     "except ImportError:",
     "if self.debug:",
+    "if sys.platform != \"win32\":",
     "pragma: no cover",
     "raise AssertionError",
     "raise NotImplementedError",
--- a/src/websockets/sync/server.py
+++ b/src/websockets/sync/server.py
@@ -2,8 +2,7 @@ from __future__ import annotations
 
 import http
 import logging
-import os
-import select
+import selectors
 import socket
 import ssl
 import threading
@@ -199,7 +198,6 @@ class WebSocketServer:
         if logger is None:
             logger = logging.getLogger("websockets.server")
         self.logger = logger
-        self.shutdown_watcher, self.shutdown_notifier = os.pipe()
 
     def serve_forever(self) -> None:
         """
@@ -214,15 +212,14 @@ class WebSocketServer:
                 server.serve_forever()
 
         """
-        poller = select.poll()
-        poller.register(self.socket)
-        poller.register(self.shutdown_watcher)
+        poller = selectors.DefaultSelector()
+        poller.register(self.socket, selectors.EVENT_READ)
 
         while True:
-            poller.poll()
+            poller.select()
             try:
                 # If the socket is closed, this will raise an exception and exit
-                # the loop. So we don't need to check the return value of poll().
+                # the loop. So we don't need to check the return value of select().
                 sock, addr = self.socket.accept()
             except OSError:
                 break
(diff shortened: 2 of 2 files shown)


aaugustin:
Thank you for doing the research. I considered using the selectors module when I wrote that code but wasn't quite sure about the benefits of the additional abstraction layer.

ceharris:
With some additional testing today, I noticed that Mac OS X doesn't return from the `select` call when the listener socket is closed. I would swear that I tested that yesterday, but I must have fooled myself. I'm going to push an additional commit to conditionally include the `os.pipe` as before when `sys.platform != win32`. With that, the only change for platforms other than Windows will be the use of `selector.DefaultSelector` instead of `select.poll`.

aaugustin:
> Mac OS X doesn't return from the select call when the listener socket is closed.

This is almost certainly why I bother with the `os.pipe()` in the first place. There are some subtle differences in how sockets behave on BSD/macOS and Linux.

ceharris:
I pushed an additional commit to restore the use of os.pipe unless the runtime platform is win32.

ceharris:
I see that the test coverage is now failing, but I'm uncertain how to address it. If you can give me a little guidance, I'll be happy to put in some time to see if I can fix it.

aaugustin:
Let's add `if sys.platform != "win32":` [between lines 63 and 64 here](https://github.com/python-websockets/websockets/blob/fe629dede6eb083013e0d2373d5c3120c0078db3/pyproject.toml#L63).

For the purposes of coverage testing, we cannot realistically get branch coverage here with a CI that runs on Linux and this isn't a sufficient reason to run coverage testing on Windows.

aaugustin:
As far as I understand, this issue makes the new sync API for servers unusable under Windows. I think this deserves a bugfix release. I can take care of backporting to the 11.x branch and making a 11.0.3 release.

ceharris:
> As far as I understand, this issue makes the new sync API for servers unusable under Windows.

This was indeed my motivation for fixing it.

ceharris:
Do you need anything else from me on this PR?

aaugustin:
Thank you, I don't need anything else. I was on holidays last week. Probably I'll do it today or tomorrow.

aaugustin:
Version 11.0.3 will be available on PyPI as soon as https://github.com/python-websockets/websockets/actions/runs/4907582069 completes.

ceharris:
Thank you!

PR #1357, opened 2023-05-18
Provide an enum for close codes.

### LINKED ISSUE #1335: Add a `CloseCode` `Enum`
I was looking for an `Enum` containing definitions of the standard close codes, and couldn't find one. Is this something that could be provided to users? I think an `IntEnum` would be a good way to implement this.

It seemed like this library would be the correct place to implement such an `Enum` for Python users.

### References

- https://www.rfc-editor.org/rfc/rfc6455.html#section-7.4.1
- https://developer.mozilla.org/en-US/docs/Web/API/WebSocket/close
- https://github.com/Luka967/websocket-close-codes
- https://learn.microsoft.com/en-us/dotnet/api/system.net.websockets.websocketclosestatus?view=net-8.0

aaugustin:
Fix #1335.

### DIFF
--- a/src/websockets/frames.py
+++ b/src/websockets/frames.py
@@ -52,48 +52,69 @@ DATA_OPCODES = OP_CONT, OP_TEXT, OP_BINARY
 CTRL_OPCODES = OP_CLOSE, OP_PING, OP_PONG
 
 
-# See https://www.iana.org/assignments/websocket/websocket.xhtml
-CLOSE_CODES = {
-    1000: "OK",
-    1001: "going away",
-    1002: "protocol error",
-    1003: "unsupported type",
+class CloseCode(enum.IntEnum):
+    """Close code values for WebSocket close frames."""
+
+    NORMAL_CLOSURE = 1000
+    GOING_AWAY = 1001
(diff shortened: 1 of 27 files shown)


johnthagen (approved):

src/websockets/frames.py
```diff
@@ -52,48 +52,69 @@ class Opcode(enum.IntEnum):
 CTRL_OPCODES = OP_CLOSE, OP_PING, OP_PONG
 
 
-# See https://www.iana.org/assignments/websocket/websocket.xhtml
-CLOSE_CODES = {
-    1000: "OK",
-    1001: "going away",
-    1002: "protocol error",
-    1003: "unsupported type",
+class CloseCode(enum.IntEnum):
```
johnthagen, line 55: I would suggest decorating this with [`@enum.unique`](https://docs.python.org/3/library/enum.html#enum.unique). This will prevent accidental aliasing or typos where the same number is used on two enum members.

aaugustin: No point burning CPU to check this every time someone imports websockets. The list is never going to change anyway.

PR #1371, opened 2023-06-14
test: allow WEBSOCKETS_TESTS_TIMEOUT_FACTOR to be float

kxxt:
It is desirable to be able to set `WEBSOCKETS_TESTS_TIMEOUT_FACTOR` to a floating point value like `1.1`.


aaugustin:
I'm not sure how `1.1` is different from `2` but, hey, this change won't hurt.

PR #1390, opened 2023-08-02
Update authentication.rst

128f:
keyword arg name changed


aaugustin:
No it didn't: https://github.com/python-websockets/websockets/blob/main/src/websockets/legacy/client.py#L436

aaugustin:
You must be confusing with: https://github.com/python-websockets/websockets/blob/main/src/websockets/sync/client.py#L139

128f:
Oh, apologies. Thanks for the clarification.

Issue #1385, opened 2023-07-20
Use `lazy_loader` for imports or implement stubs/hints in `lazy_import`


wolph:
Most IDE's currently [don't properly understand](https://websockets.readthedocs.io/en/stable/faq/misc.html#why-does-my-ide-fail-to-show-documentation-for-websockets-apis
) the example code due to the current implementation of the `lazy_import` function. Luckily, there's a simple solution for this besides changing your imports from `import websockets` to `import websockets.client` :)

There is a great little project called `lazy_loader` that effectively does the same as what `lazy_import` does now but it also properly supports type hints and stubs: https://github.com/scientific-python/lazy_loader#type-checkers

I believe a switch to `lazy_loader`, or an update for `lazy_import` if you want to avoid a dependency, could improve the usability of the library quite a bit.

128f:
After reading this I was curious how this project gets built, and eventually went looking for signs of a dependency management tool, but I couldn't find one.

Then I found out [It doesn't have any dependencies.](https://github.com/python-websockets/websockets/blob/adfb8d69a7a1f6f4c8381c9e7182619d202c3cf2/docs/intro/index.rst#requirements), which is rather incredible.

Given that, I would guess the maintainer of this project would be reluctant to add an outside project as a dependency.

aaugustin:
I can confirm that I'm reluctant to add dependencies. Indeed, dependencies increase the cost of maintenance — simply because they increase the number of things that can change outside of my control — while I cannot dedicate much time to maintaining websockets.

I looked at `lazy_loader` when this issue was filed. It appears to require duplicating type information in `.pyi` files, which I don't want to do.

I think I'm just going to adopt explicit import paths everywhere eventually. Convenience imports were cool when the library was younger. We've outgrown them 🤷

aaugustin:
PS: why would websockets need any dependencies? ;-)

wolph:
I totally get that :)
Many of my libraries have no dependencies or only dependencies I maintain myself. Less stuff that can break that way ;)

As a 3rd option, I believe that a stub file could fix it as well without requiring any other changes. The only issue with stub files is that they could become out of date. But this library is rather stable and you could easily have pyright automatically test for that.

aaugustin:
439dafa6 should fix it. Duplicating 50 lines in a file is still easier to maintain than taking a dependency.

PR #1400, opened 2023-09-08
fix: add `WebSocketClientProtocol` and `connect` to all for linters

Matthew-Burkard:
With this linters and mypy work better with websockets when importing these such as:
```
from websockets.client import connect, WebSocketClientProtocol
```

### DIFF
--- a/src/websockets/client.py
+++ b/src/websockets/client.py
@@ -41,7 +41,7 @@ from .utils import accept_key, generate_key
 from .legacy.client import *  # isort:skip  # noqa: I001
 
 
-__all__ = ["ClientProtocol"]
+__all__ = ["ClientProtocol", "WebSocketClientProtocol", "connect"]
 
 
 class ClientProtocol(Protocol):


aaugustin:
Thank you for the proposal. When you say "work better", could you clarify what effect you observed?

Overall, it looks like the right thing to do here, but I'm curious to understand what it changes exactly.

If I adopt this solution, I will want to do it across all modules where a similar pattern exists.

Matthew-Burkard:
If you import `websockets` and call `websockets.connect` like so:

```python
import websockets

async def main() -> None:
    websocket = await websockets.connect("ws://localhost:8000/")
```

mypy raises the following error:
```
error: Module has no attribute "connect"  [attr-defined]
Found 1 error in 1 file (checked 1 source file)
```

If to avoid this you import `connect` from `websockets.client` as such:

```python
from websockets.client import connect

async def main() -> None:
    websocket: WebSocketClientProtocol = await connect("ws://localhost:8000/")
    await websocket.close()
```

PyCharm builtin linter check for "Accessing a protected member of a class or a module" will flag that import since `connect` is not in `__all__` for `websockets.client` module.

I want to add `connect` to `__all__` for this reason and `WebSocketClientProtocol` for type annotation purposes, e.g.
```python
websocket: WebSocketClientProtocol = ...
```

Doing this, in addition to satisfying the linter, enables auto-completions and go to source functionality within PyCharm.

aaugustin:
This looks like an elegant and cheap solution to the issue described in #1385 (and several earlier issues). I'll look into it!

aaugustin:
I fixed it with a slightly different approach in all modules where the issue could occur. I don't have PyCharm so I cannot confirm the fix.

aaugustin:
(It doesn't help mypy understand `websockets.connect` — `connect` was already in `websockets.__all__`.)

PR #1406, opened 2023-10-06
Allow plain integers as HTTP status codes

Rosuav:
Since an IntEnum is largely equivalent to the corresponding integer, allow HTTP status codes to be specified as (eg) `200` rather than `http.HTTPStatus.OK`.

### DIFF
--- a/src/websockets/legacy/server.py
+++ b/src/websockets/legacy/server.py
@@ -217,6 +217,7 @@ class WebSocketServerProtocol(WebSocketCommonProtocol):
                         ),
                     )
 
+                status = http.HTTPStatus(status)
                 headers.setdefault("Date", email.utils.formatdate(usegmt=True))
                 if self.server_header is not None:
                     headers.setdefault("Server", self.server_header)
--- a/tests/legacy/test_client_server.py
+++ b/tests/legacy/test_client_server.py
@@ -543,6 +543,17 @@ class CommonClientServerTests:
         with contextlib.closing(response):
             self.assertEqual(response.code, 200)
 
+    class ProcessRequestOKServerProtocolNumeric(WebSocketServerProtocol):
+        async def process_request(self, path, request_headers):
+            return 200, [], b"OK\n"
+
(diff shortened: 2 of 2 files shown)


aaugustin:
Quick review notes:

- Test needed
- No documentation needed
- What about other implementations?

Rosuav:
Added a test, but either I'm running the tests wrongly, or there are other failures. Can someone check please? I'm getting failures in eleven tests, mostly `ImportError: cannot import name 'CloseCode' from 'websockets.frames'`, when running tests on the main branch.

Other implementations? If you mean the fact that I put this into the `legacy/` directory, that seems to be where this happens, regardless of how it's called upon. My original code (where I ran into this problem) uses `async with websockets.serve(..., process_request=...)` and has that function asynchronously return the three-tuple.

aaugustin:
Thank you for the test case :-)

My comment about "other implementations" was a note to myself to check the Sans-I/O and the threading implementation because they're probably vulnerable to the same issue.

aaugustin:
Other implementations: actually, websockets already contains similar code in the Sans-I/O implementation.

This supports aligning the legacy implementation to this behavior, like you've done.

aaugustin:
The test that you added passes without code changes. The conversion from `int` to `HTTPStatus` is happening here:

https://github.com/python-websockets/websockets/blob/f62d44ac1a0a606fe5753cf0017ce9e1dbf3640e/src/websockets/exceptions.py#L338

aaugustin:
And there is already a test for this behavior: https://github.com/python-websockets/websockets/blob/f62d44ac1a0a606fe5753cf0017ce9e1dbf3640e/tests/legacy/test_client_server.py#L766-L772

Rosuav:
Are you sure it all works? I wasn't running into issues with a type checker, but with an AttributeError when it tried to get some piece of information from the enum (which wasn't on the plain int).

aaugustin:
I didn't work in the last release. It works in the release I made on Saturday.

PR #1439, opened 2024-02-07
Fix wrong RFC number

MtkN1:
While studying the websockets source code, I found an incorrect RFC number in the comments and test docstrings.

- Incorrect: [RFC 6544](https://datatracker.ietf.org/doc/html/rfc6544) TCP Candidates with Interactive Connectivity Establishment (ICE)
- Correct: [RFC 6455](https://datatracker.ietf.org/doc/html/rfc6455) The WebSocket Protocol

I did this keyword search and fixed everything.

I truly appreciate your efforts in maintaining this fantastic library!


aaugustin:
Thank you for the fix!

Issue #1419, opened 2023-11-28
`process_request` returning response runs connection handler


Flowrey:
I'm using:
* Python: 3.12
* websockets: 12.0

When using the threading implementation like in this code:
```python
import http
from websockets.sync.server import serve, ServerConnection, Request, Response
from websockets.http11 import datastructures

def health_check(websocket: ServerConnection, request: Request):
    if request.path == "/healthz":
        return Response(http.HTTPStatus.OK, "OK", datastructures.Headers([]), b"OK")
    else:
        return None


def echo(websocket):
    for message in websocket:
        websocket.send(message)

def main():
    with serve(echo, "localhost", 8765, process_request=health_check) as server:
        server.serve_forever()

main()
```
Any call to the /healthz endpoint works but raise a `ConnectionCloseError`:
```
connection handler failed
Traceback (most recent call last):
  File "/home/flowrey/Code/poc/venv/lib64/python3.12/site-packages/websockets/sync/server.py", line 499, in conn_handler
    handler(connection)
  File "/home/flowrey/Code/poc/server.py", line 13, in echo
    for message in websocket:
  File "/home/flowrey/Code/poc/venv/lib64/python3.12/site-packages/websockets/sync/connection.py", line 162, in __iter__
    yield self.recv()
          ^^^^^^^^^^^
  File "/home/flowrey/Code/poc/venv/lib64/python3.12/site-packages/websockets/sync/connection.py", line 201, in recv
    raise self.protocol.close_exc from self.recv_events_exc
websockets.exceptions.ConnectionClosedError: no close frame received or sent
```

MauroAntonino:
I noticed that it is receiving empty bytes as a data response, and this causes the socket to close.
https://github.com/python-websockets/websockets/blob/e217458ef8b692e45ca6f66c5aeb7fad0aee97ee/src/websockets/sync/connection.py#L549-L550
https://github.com/python-websockets/websockets/blob/e217458ef8b692e45ca6f66c5aeb7fad0aee97ee/src/websockets/sync/connection.py#L614-L616

Another thing is that when the status code is not 101, the state does not end up being OPEN, which then closes the connection as well.
https://github.com/python-websockets/websockets/blob/e217458ef8b692e45ca6f66c5aeb7fad0aee97ee/src/websockets/server.py#L541-L547
[...]

PR #1443, opened 2024-02-21
Verify if is a websocket connection

MauroAntonino:
Now, every HTTP request is handled as if it were a WebSocket connection. However, when an HTTP request is made and is expected to close the connection, an error arises because it is not supposed to receive an HTTP connection in the handler.

This PR verifies if it is actually a WebSocket connection before handling it.

### DIFF
--- a/src/websockets/sync/server.py
+++ b/src/websockets/sync/server.py
@@ -516,14 +516,8 @@ def serve(
             return
 
         try:
-            protocol.logger.error(connection.__dict__.get('request').headers
-                                  , exc_info=True)
-            protocol.logger.error(is_websocket(connection)
-                                  , exc_info=True)
             if is_websocket(connection):
                 handler(connection)
-            else:
-                connection.close()
         except Exception:
             protocol.logger.error("connection handler failed", exc_info=True)
             connection.close(CloseCode.INTERNAL_ERROR)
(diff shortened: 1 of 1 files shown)


aaugustin:
Looks related to #1419.

davidawesome02:
I may be wrong; but according to [mozilla](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Upgrade) the upgrade header can be used for http requests, not just switching to websockets, so it should be explicitly checked for the upgrade to websocket.

This feature seams to be generally unused and depreciated, so in most cases should not matter.

MauroAntonino:
Yes, you are right. It must check if the header has the websocket string explicitly. I made a commit to fix right now.

aaugustin:
Can you provide more information about what you're doing, what you're expecting, and what you're getting?

> when an HTTP request is made and is expected to close the connection

In this case, the call to `connection.handshake(...)` — ten lines above the code that you changed — is expected to raise an exception. As a consequence, that code shouldn't be reached.

If this doesn't work as expected, the fix consists in making `connection.handshake(...)` raise an exception, because that's the method in charge of telling if this is a WebSocket connection or not.

We mustn't reimplement some of the protocol logic outside of the Sans-I/O layer.

aaugustin:
(And sorry for the late answer — I haven't found time to work on this project in the recent months.){

aaugustin:
(And sorry for the late answer — I haven't found time to work on this project in the recent months.)

aaugustin:
Superseded by #1500.

PR #1444, opened 2024-02-22
Disallow cross-origin redirect when host, port, path or sock is set. (#631)

### LINKED ISSUE #631: Don't follow redirects when path, sock, host or port are overridden
(from https://github.com/aaugustin/websockets/pull/544#issuecomment-505837262)

Since #528 websockets follows redirects.

I believe that this feature is incompatible with connection overrides (either with host + port or with sock). If you need to control exactly where you're connecting, I assume you aren't expecting to be redirected somewhere else.

This deserves documentation at a minimum, perhaps a warning or even an exception.

SLiV9:
This fixes #631.

### DIFF
--- a/src/websockets/legacy/client.py
+++ b/src/websockets/legacy/client.py
@@ -515,24 +515,30 @@ class Connect:
             loop=_loop,
         )
 
+        host: Optional[str] = kwargs.pop("host", None)
+        port: Optional[int] = kwargs.pop("port", None)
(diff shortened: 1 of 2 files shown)


aaugustin:
I'm not going to fix this in the legacy implementation.

#1495 gets it right in the new asyncio implementation.

It's also a bit smarter e.g. it supports same-origin redirects when host and port are overridden, which is OK. This PR breaks that use case, which could be considered as a regression by users.

PR #1448, opened 2024-02-28
docs: Correct the example for "Starting a server" in the API reference

xuanzhi33:
It seems that `server.close()` is not an `async` function, so I changed
```python
await server.close()
```
to
```python
server.close()
await server.wait_closed()
```
in the API reference.

### DIFF
--- a/docs/faq/server.rst
+++ b/docs/faq/server.rst
@@ -300,7 +300,8 @@ Here's how to adapt the example just above::
 
         server = await websockets.serve(echo, "localhost", 8765)
         await stop
-        await server.close(close_connections=False)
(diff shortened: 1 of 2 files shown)


aaugustin:
Thank for you the fix — indeed close() isn't a coroutine: https://github.com/python-websockets/websockets/blob/e217458ef8b692e45ca6f66c5aeb7fad0aee97ee/src/websockets/legacy/server.py#L724

I'm going to check how I introduced this mistake and merge the fix.

xuanzhi33:
I found the same mistake in the FAQ and fixed it.

PR #1454, opened 2024-03-25
Fix the issue of redirecting from ws to wss.

xianglin1998:
I found that when configuring our company's Apache server, redirecting ws requests to wss and using this websocket library for connection would result in the error "The plain HTTP request was sent to HTTPS port". After using Wireshark for packet capture, I found that this library did not properly configure SSL. After DEBUG, I confirmed that the problem occurred in the 'handle_redirect' function: 'create_connection' requires the correct use of the SSL keyword.

### DIFF
--- a/src/websockets/legacy/client.py
+++ b/src/websockets/legacy/client.py
@@ -573,6 +573,10 @@ class Connect:
                 *factory.args,
                 **dict(factory.keywords, host=new_wsuri.host, port=new_wsuri.port),
             )
+            # Fix the issue of "The plain HTTP request was sent to HTTPS port" 
+            # caused by forgetting to enable SSL after redirecting from ws to wss.
+            if new_wsuri.secure:
+                self._create_connection.keywords["ssl"] = True
             # Replace the host and port argument passed to create_connection.
             self._create_connection = functools.partial(
                 self._create_connection.func,


owl-with-beard:
Push for visibility. I spent a day to figure out the exact same thing as @xianglin1998 .
If the server requires SSL on the redirected address, the redirection fails as the connection to the new address is still made using {'ssl': False}.

The proposed commit fixes the problem, as it adjusts the handle_redirect function to account for WS --> WSS redirects.
However, I think you have to update the protocol factory too (the lines right above the proposed commit)

```
factory = self._create_connection.args[0]
factory = functools.partial(
   factory.func,
   *factory.args,
   **dict(factory.keywords, host=new_wsuri.host, port=new_wsuri.port, secure=new_wsuri.secure)
)
```

(The other way, WSS --> WS is already correcly addressed by raising an error.)

Is there something I can do to make this pull request merge-able ?

Greetings

aaugustin:
This patch breaks redirects from a `wss://` to another `wss://` URI when the `ssl` argument of `connect` is set to a value other than `True` because it replaces the custom `SSLContext` by a default context. If you run the tests, you get an error for this reason.

xianglin1998:
So does the latest release fix the ws to wss redirection bug?

aaugustin:
It doesn't because it predates this pull request.

I'm working on a correct fix for this issue as we speak.

aaugustin:
Now fixed in the main branch. The fix will be included in the next release.

aaugustin:
Thank you for raising this issue!

xianglin1998:
Good！！！Thank you for your work.

PR #1469, opened 2024-05-20
Correct handle exceptions in `legacy/broadcast`

ABCurado:
When enabling raise_exceptions in the broadcast method the code will always return an exception without this change


aaugustin:
Test added in 96d3adf.

PR #1476, opened 2024-07-02
really drop python <=3.7 support

kloczek:
Filetrer all cod over `pyupgrade --py38-plus`.

### DIFF
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -8,8 +8,8 @@ import datetime
 import importlib
 import inspect
 import os
-import subprocess
 import sys
+
 
 # -- Path setup --------------------------------------------------------------
 
@@ -25,7 +25,6 @@ project = "websockets"
 copyright = f"2013-{datetime.date.today().year}, Aymeric Augustin and contributors"
 author = "Aymeric Augustin"
 
-from websockets.version import tag as version, version as release
 
 
 # -- General configuration ---------------------------------------------------
@@ -79,6 +78,7 @@ autodoc_typehints_description_target = "documented"
 # Workaround for https://github.com/sphinx-doc/sphinx/issues/9560
 from sphinx.domains.python import PythonDomain
 
+
 assert PythonDomain.object_types["data"].roles == ("data", "obj")
 PythonDomain.object_types["data"].roles = ("data", "class", "obj")
 
@@ -96,6 +96,7 @@ exclude_patterns = ["_build", "Thumbs.db", ".DS_Store"]
 
 # Configure viewcode extension.
 from websockets.version import commit
+
 
 code_url = f"https://github.com/python-websockets/websockets/blob/{commit}"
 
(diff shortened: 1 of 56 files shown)


docs/conf.py
```diff
@@ -25,7 +25,6 @@
 copyright = f"2013-{datetime.date.today().year}, Aymeric Augustin and contributors"
 author = "Aymeric Augustin"
 
-from websockets.version import tag as version, version as release
```
aaugustin, line 28: That will break generating the docs.

example/django/notifications.py
```diff
@@ -7,6 +7,7 @@
 import django
 import websockets
 
+
```
aaugustin, line 10: Please avoid gratuitous formatting changes. Also, probably the autoformatter built into the project would revert it at the next commit.

example/echo.py
```diff
@@ -1,8 +1,10 @@
 #!/usr/bin/env python
 
 import asyncio
+
 from websockets.server import serve
 
+
```
aaugustin, line 7: Again, gratuitous change that will make the docs look worse.

aaugustin (changes requested):
Some of the changes are valid — e.g. unnecessary imports — but there's also many incorrect changes. Unfortunately it will probably be faster for me to redo the changes than to review and guide you over GitHub comments :-/ I gave up reviewing after a few files because the signal / noise ratio wasn't great.

aaugustin:
I think I'm going to redo individual changes — again, faster to redo than review — which will eventually lead to the result that you're aiming for. Thank you for the sugestion! That's a good opportunity to clean up.

kloczek:
OK so just cannel this PR and do that your way.
If you see some code wrongly filtered please discuss that with `pyupgrade` maintainer.
FYI I've filtered `websockets` code up to --py310-plus and it passes test suite and I don't see any issues on interaction with other modules. code.

aaugustin:
Union types were introduced in Python 3.10. See https://peps.python.org/pep-0604/. As a consequence, this change isn't compatible with Python 3.8.

```
$ python3.8
Python 3.8.16 (default, Apr  2 2023, 07:49:54)
[Clang 14.0.0 (clang-1400.0.29.202)] on darwin
Type "help", "copyright", "credits" or "license" for more information.
>>> def foo(bar=str | None): pass
...
TypeError: unsupported operand type(s) for |: 'type' and 'NoneType'
>>>
```

```
$ python3.10
Python 3.10.10 (main, Apr  2 2023, 07:49:30) [Clang 14.0.0 (clang-1400.0.29.202)] on darwin
Type "help", "copyright", "credits" or "license" for more information.
>>> def foo(bar=str | None): pass
...
>>>
```

aaugustin:
Oops. Actually not a problem thanks to `from __future__ import annotations`. I had missed that.

aaugustin:
Need to check that the docs still generate properly.

aaugustin:
As of f45286b all the good changes from this PR are in the main branch.

Issue #1532, opened 2024-10-29
ServerConnection._aiter_ not raising websockets.execptions.ConnectionClosed


Electron4444:
I purpose a slight change to the ServerConnection._aiter_ method, to raise the ConnectionClosed Exception on excepting the CloseConnectionOK Exception. This would lead to the funtionality documented in Patterns for the Producer to be true for the Consumer as well and enable the possibility for a server waiting on receiving messages to close the handler for the corresponding client more easily on a connection close

aaugustin:
I don't want to do this for two reasons:

1. I think that the current API is the right one. Iterating until the end of a connection closed normally isn't a problem, therefore it shouldn't raise an exception.
2. It's a significant backwards-incompatibility: it will trigger an exception in an extremely common code path. Even if it was an improvement, probably I wouldn't make the change for that reason.

Electron4444:
Could you maybe add a keyword option to set which enables this. I'm gonna make a pull- request containing the proposal. I would just let this controlling optional parameter default to None but have it raise the exception if enabled. This would make handeling Connection Closes inside the handler function provided to serve way easier.

aaugustin:
No. Good API design is about providing a good API, not about providing knobs to get every possible behavior.

The problem with the change that you're proposing is that a reader of the code cannot tell anymore what `for message in websocket: ...` does. The reader needs to look at where the websocket object is defined. That's bad.

PR #1534, opened 2024-11-01
Implementationn of optional Solve for Suggested Change in Issue #1532

Electron4444:
This adds the optional keyword `raise_on_close` to the Constructor of the serve class allowing this parameter to propogate to the ServerConnection object, where it allows the following to raise an exception over the iteration:

```
async for message in server_connection:
    # handle Message
```
This enables the handeling of Connection Closure inside the Connection handeling function and can save the rescources of a handler continuing to run asynchronously without any Connection still being open. Especially for large scale Project this paramter could save a lot of rescources in RAM and of the CPU as the handlers could be easily freed from these resources upon the connection closing. This could be done in the following fashion:
```
try:
    while True:
        async for message in server_connection:
            # handle Message
except websockets.exception.ConnectionClosed:
    server_connection.close()
    return # return from handler 
```
This could simplify the handeling process all just by adding the feature: `serve([handler_function],[host],[port],[*args],[**kwargs],raise_on_close=True)` where `[]` are just placeholders here.

In this manner the async iteration would now behave similar to recv and send, if this behaviour is wanted, which gives a flexible usage for the async iteration and doesn't break systems designed for earlier versions. It does however break backwards compatibility because of the additional keyword parameter. In the future this might be resolved by adding a keyword garbage at the end in the Constructor to allow future non-breaking changes more easily.

### DIFF
--- a/src/websockets/asyncio/connection.py
+++ b/src/websockets/asyncio/connection.py
@@ -57,6 +57,7 @@ class Connection(asyncio.Protocol):
         close_timeout: float | None = 10,
         max_queue: int | tuple[int, int | None] = 16,
         write_limit: int | tuple[int, int | None] = 2**15,
+        raise_on_close: bool | None = None
     ) -> None:
         self.protocol = protocol
         self.ping_interval = ping_interval
@@ -132,6 +133,8 @@ class Connection(asyncio.Protocol):
         self.drain_waiters: collections.deque[asyncio.Future[None]] = (
             collections.deque()
         )
+
+        self.raise_on_close: bool = raise_on_close is not None
 
     # Public attributes
 
@@ -216,6 +219,8 @@ class Connection(asyncio.Protocol):
             while True:
                 yield await self.recv()
         except ConnectionClosedOK:
+            if self.raise_on_close:
+                raise ConnectionClosed(None,None)
             return
 
     async def recv(self, decode: bool | None = None) -> Data:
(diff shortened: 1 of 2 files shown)


Electron4444:
If this request is denied, I would possibly purpose adding a new method or property that returns an asnychronus iterator with this behaviour. This would still support backwards compatibility and would allow for the proposed handleing optimization.
If you want me to implement this instead, I already have an idea. Please do reach out in that case, as I would really look forward to contribute 😊

aaugustin:
> I would really look forward to contribute 😊

Look, you didn't make an effort to understand why things are the way they are and you steamroll a change that's clearly a bad idea, why paying nowhere near the level of attention expected from contributors to this project: the first word in your commit message has a typo!

Please find another project to contribute to. Thank you.

Electron4444:
Thanks for your feedback, even though your criticism was clearly not constructive in the end.
I do get your point now and see why it was, in fact, a bad addition.
I have never contributed before and tried to find somewhere to start.
The exaggeration of my grammatical errors in the titles isn't nice, though.

But thanks for taking the time to review it. I'll try to improve.

Electron4444:
Thanks for your feedback, even though your criticism was clearly not constructive at the end.
I do get your point now and see why it was, in fact, a bad addition.
I have never contributed before and tried to find somewhere to start.
The exaggeration of my grammatical errors in the titles isn't nice, though.

But thanks for taking the time to review it. I'll try to improve.
If you do answer again, I'd really appreciate it if you could give me some tips on a good way to get started with contributing to open-source projects.
Thanks again.

PR #1539, opened 2024-11-11
Add missing import

simontorres:
When handling the ConnectionClosedOK exception the tutorial didn't mention the need to import websockets or just the exception

### DIFF
--- a/docs/intro/tutorial1.rst
+++ b/docs/intro/tutorial1.rst
@@ -271,6 +271,8 @@ spot real errors when you add functionality to the server. Catch it in the
 
 .. code-block:: python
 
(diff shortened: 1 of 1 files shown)


aaugustin:
Good catch - regression from 14ca557f53cf19084eb64aef2e4563e5630c211b that changed `import websockets; websockets.serve` to `from websockets.asyncio.server import serve; serve`

aaugustin:
I fixed it a bit differently, by using the real import path rather than the shortcut.

Also I fixed a few other instances of this problem at the same time.

aaugustin:
Thank you for reporting this problem!

PR #1571, opened 2025-01-09
Aligns `Connection.recv(timeout=0)` with docstring

### LINKED ISSUE #1552: Unclear behaviour of recv(timeout=0) for Server (threading)
I think the behavior of recv(timeout=0) for the Thread-based Server is unclear. 
Due to the design of my application, I can't use the "for message in websocket" and have to resort to an update() function that I call frequently to check if messages were received: 

```

def update(self) -> None:
	"""
	Check for received data with proper handling of timeouts and connection closure.
	"""
	while True:
		try:
			# Attempt to receive data with a timeout
			data_rx = self.ws_client.recv(timeout=0)
		except TimeoutError:
			# No data received within the timeout period
			break
		except websockets.exceptions.ConnectionClosed as e:
			# Handle connection closure
			Log(f"Connection closed: {e}")
			self._is_connected = False
			break
		except Exception as e:
			# Handle any other exceptions
			Log(f"Error occurred: {e}")
			self._is_connected = False
			break

		# If no data received, exit the loop
		if data_rx is None:
			break

		self.log(f"RX: {data_rx}")


```

I was expecting when I set timeout=0 that only when a message was received I get it otherwise TimeoutError is raised so it's not blocking. 
the behavior I'm getting is that with very low values like timeout=0.001 (1ms) it works, but with 0 I don't receive any messages. 

Have I understood timeout=0 incorrectly? In the documentation it says: 
"Set timeout to 0 to check if a message was already received." [see](https://websockets.readthedocs.io/en/stable/reference/sync/server.html#websockets.sync.server.ServerConnection.recv)

mattyoungberg:
Fix #1552

NOTE: I added a couple of tests, but one of the two is still not passing. My intention is to fix and then squash the fix into this commit. At which point, this note will be removed.

The docstring for `Connection.recv` suggests that if you provide the argument `timeout=0`, you can retrieve any message in waiting:

"Set `timeout` to `0` to check if a message was already received."

Current behavior is that this does nothing and instead makes a `TimeoutError` immediately throw on the first use of a derived `Deadline`.

However, being able to check for a full message without blocking unneccesarily (aka, waiting for the server to send a message) would be useful, especially if your websocket client needs to be able to continue to do work in another thread and so periodically ceding control is important.

Most of the problem comes from the fact that the code doesn't treat a timeout value of 0 as special, and the code path will create a `Deadline` with it, and then rely on the side effects of blocking network I/O and exceptions in order to guarantee the contract of `Connection.recv()` which is to ultimately deliver a full message.

The error has been addressed by allowing the API of `Assembler.get()` to potentially return `None` when a new param, `peek`, is set to True (the default is False and maintains previous behavior). The intention of this param is to allow the method to acquire the class's mutex, check if there are any frames in the `Assembler` instance, and if not, return `None`. The only place where this param is used is within `Connection.recv()`, which is now able to treat the exceptional case of `timeout=0` by setting up it's arguments to `Assembler.get()` without otherwise changing its code path.

Tests:

- Two tests were added: one to make sure that a client gets back `None` when `timeout=0` and the underlying frames queue is empty, and another that returns the correct message when the underlying frames queue has a frame.

### DIFF
--- a/src/websockets/sync/connection.py
+++ b/src/websockets/sync/connection.py
@@ -9,7 +9,7 @@ import threading
 import uuid
 from collections.abc import Iterable, Iterator, Mapping
 from types import TracebackType
-from typing import Any
+from typing import Any, Optional
 
 from ..exceptions import (
     ConcurrencyError,
@@ -216,11 +216,14 @@ class Connection:
         """
         try:
             while True:
-                yield self.recv()
+                # can't receive None if peek=False (default)
+                yield self.recv()  # type: ignore[misc]
         except ConnectionClosedOK:
             return
 
-    def recv(self, timeout: float | None = None, decode: bool | None = None) -> Data:
+    def recv(
+        self, timeout: float | None = None, decode: bool | None = None
+    ) -> Optional[Data]:
         """
         Receive the next message.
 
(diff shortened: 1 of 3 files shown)


mattyoungberg:
Closing in favor of #1573

PR #1575, opened 2025-01-11
Add regex support in `ServerProtocol(origins=...)`

daH005:
Hello. In my project I tried to pass a regular expression to `origins`, but it didn't work.
I think this will not be superfluous for the project.
Idea taken from [https://github.com/corydolphin/flask-cors/blob/main/flask_cors/core.py](https://github.com/corydolphin/flask-cors/blob/main/flask_cors/core.py)

### DIFF
--- a/src/websockets/asyncio/server.py
+++ b/src/websockets/asyncio/server.py
@@ -4,6 +4,7 @@ import asyncio
 import hmac
 import http
 import logging
+import re
 import socket
 import sys
 from collections.abc import Awaitable, Generator, Iterable, Sequence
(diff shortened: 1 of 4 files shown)


aaugustin:
Yes, that's a valid improvement.

I don't like the heuristics to tell what's likely a regex. Instead, how about we keep strict backwards-compatibility when the argument is a `str` and add support for `re.Pattern`?

daH005:
I agree with you. I did this and I also added tests.

aaugustin:
I don't always find time to work on websockets outside of weekends — but I'll get to your MR eventually. Thank you.

aaugustin:
If you want to run the same checks as CI locally, you can simply run `tox` (and ignore Python version that aren't installed — this MR won't break on specific Python versions). Or you can use `make style` (fixes automatically formatting, imports, etc.) +`make coverage` (runs unittest) + `make types` (checks types).

PR #1590, opened 2025-02-04
Fix incorrect asyncio import in README

floryst:
The primary README has an invalid import in the asyncio example.

Changes:
- `serve` is now correctly imported from `websockets.asyncio.server`


aaugustin:
I fixed it in all Python files and missed in in the README 🤦‍♂️ Thank you.

PR #1599, opened 2025-02-25
Gracefully closing the cli when KeyboardInterrupt signal is sent 

a-alak:
While setting up entry point for the cli (#1598) we saw that the cli does not properly close the connection on KeyboardInterrupt. 

Doing a bit of research it seems that asyncio will catch KeyboardInterrupt outside of the task and then cancel the task. From python 3.11 asyncio will raise a CancelledError in the task instead. Catching this in our tasks enable the cli to gracefully exit with no error. In python 3.9 and 3.10 this will still properly close the websocket connection, but you will get a KeyboardInterrupt error (still outside our task, so catching this error inside the task does not change anything). 

Should we also silent the error in python 3.9 and python 3.10?

### DIFF
--- a/src/websockets/cli.py
+++ b/src/websockets/cli.py
@@ -2,9 +2,9 @@ from __future__ import annotations
 
 import argparse
 import asyncio
-from asyncio import CancelledError
 import os
 import sys
+from asyncio import CancelledError
 from typing import Generator
 
 from .asyncio.client import ClientConnection, connect
--- a/src/websockets/cli.py
+++ b/src/websockets/cli.py
@@ -2,6 +2,7 @@ from __future__ import annotations
 
 import argparse
 import asyncio
+from asyncio import CancelledError
 import os
 import sys
 from typing import Generator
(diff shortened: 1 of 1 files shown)


aaugustin:
Ah, I should have tested it, I'm getting the same 🤦‍♂️ Looks like I always exit with ^D (just a habit to exit an interactive shell) and never with ^C so I didn't notice. Previously there was a 10s delay; now there's a traceback; we're almost there :-)

aaugustin:
Thank you, this is the correct fix indeed. I added more comments and committed the fix.

a-alak:
Amazing! Happy to help 🙌🏽 Thank you for quick response! 
Just curious, how is your release cycle? Looking forward to run the cli with uvx 😇

aaugustin:
I just released 15.0.1. It will be on PyPI as soon as https://github.com/python-websockets/websockets/actions/runs/13684207641 completes.

a-alak:
Amazing! Thanks 🙏🏽

PR #1605, opened 2025-03-16
chore: correct time diff sign in compression doc

serozhenka:
Just a small compression doc correction


aaugustin:
Good catch, thank you!

Issue #1488, opened 2024-08-22
Threading server should close connections on exit


aaugustin:
Essentially, it should behave like the asyncio server. It could also gain the connections attribute.

aaugustin:
This would also be more elegant than f9cea9cca568dc92704de3744639eb4248278a8f.

aaugustin:
Implementing this may hit #1596.

TobiasKaiser:
I have the same / a related issue: To terminate a threading websockets server with open connections, I have to hit Ctrl+C twice. This is a minor annoyance, but I would like to fix it. For my use case, setting daemon=True for threading.Thread in Server.serve_forever would fix the problem, as all daemon threads are automatically terminated when the main thread terminates.

Maybe someone can elaborate on the comment in sync/server.py: "we cannot use daemon threads, or else all connections would be terminate brutally when closing the server." Is there currently a way to terminate connection threads less brutally than daemon=True would? Or is there a specific use case where it is desired to terminate the listener thread but keep connection threads running?

aaugustin:
Tobias - yes, I want websockets to support draining connections i.e. the server no longer accepts connections, then two options:

* existing connections are closed — this should lead handlers to terminate, and they should have time to clean up
* existing connections are left open and we just wait until all clients disconnect

TobiasKaiser:
Thanks! This will simplify my code 👍

PR #1615, opened 2025-04-06
Implement connection tracking and graceful shutdown (sync server)

darkhaniop:
Hi again, I implemented the connection tracking and graceful server shutdown procedure that I explained in [issue #1488](https://github.com/python-websockets/websockets/issues/1488#issuecomment-2778939350).

### Server shutdown behavior

In the current version, the sync server does not have a way of tracking active connections. After launching the handler into a separate thread, the Server instance forgets about that connection. When we want to interrupt the server process with an interrupt, this can cause the process to continue running due to remaining active connections.

I updated `src/websockets/sync/server.py` to add logic for better handling graceful shutdowns. I also added optional arguments `code` and `reason` that are passed to the clients with `ServerConnection.close(code, reason)` which allows additional information to be provided to the clients.

### Considerations

One of the things I considered when choosing how to update active connections is the existing API. I am not sure if there are users who instantiate Server() differently in their codebases, as opposed to the recommended `server = serve(socket, handle, ...)` approach. Therefore, I think there could be complaints about changing the signature of `conn_handler()`. So, I kept it unchanged. I did add an optional argument to the constructor of `Server` (as a keyword-only arg, so I think it won't be a problem either).

I added some tests in `tests/sync/test_server.py`, and tried to match the style of existing tests there.

Please let me know if anyone has suggestions on this.

*Update: fixed a typo "sync server does have" -> "sync server does **not** have"*

### DIFF
--- a/src/websockets/sync/server.py
+++ b/src/websockets/sync/server.py
@@ -230,6 +230,10 @@ class Server:
         logger: Logger for this server.
             It defaults to ``logging.getLogger("websockets.server")``.
             See the :doc:`logging guide <../../topics/logging>` for details.
+        connections: A set of open :class:`ServerConnection` instances
+            maintained by `handler`. When omitted, e.g., if the handler does
+            not maintain such a set, this defaults to an empty set and the
+            server will not attempt to close connections on shutdown.
 
     """
 
@@ -293,21 +297,11 @@ class Server:
             thread = threading.Thread(target=self.handler, args=(sock, addr))
             thread.start()
 
-    def shutdown(
-        self, *, code: CloseCode = CloseCode.NORMAL_CLOSURE, reason: str = ""
-    ) -> None:
+    def shutdown(self) -> None:
         """
         See :meth:`socketserver.BaseServer.shutdown`.
 
-        Shuts down the server and closes existing connections. Optional arguments
-        ``code`` and ``reason`` can be used to provide additional information to
-        the clients, e.g.,::
-
-            server.shutdown(reason="scheduled_maintenance")
-
-        Args:
-            code: Closing code, defaults to ``CloseCode.NORMAL_CLOSURE``.
-            reason: Closing reason, default to empty string.
+        Shuts down the server and closes existing connections.
 
         """
         self.socket.close()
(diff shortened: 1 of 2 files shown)


darkhaniop:
BTW, here's a script I used to check this behavior:
```python
import threading
import time
from websockets.sync.server import serve, Server

class SharedRef:
    server: Server

def server_target(shared_ref: SharedRef):

    def echo(websocket):
        for message in websocket:
            websocket.send(message)

    with serve(echo, "localhost", 8765) as server:
        shared_ref.server = server
        server.serve_forever()

def main():
    shared_ref = SharedRef()
    server_thread = threading.Thread(target=server_target, args=(shared_ref,))
    server_thread.start()
    try:
        while True:
            time.sleep(1)
    except KeyboardInterrupt:
        pass

    if shared_ref.server is not None:
        # shared_ref.server.shutdown(reason="scheduled_maintenance")
        shared_ref.server.shutdown()

if __name__ == "__main__":
    main()
```
In my test, when this server has active connections (with the updated library), KeyboardInterrupt terminates the process without delay or exceptions.

aaugustin:
> I also added optional arguments code and reason that are passed to the clients with ServerConnection.close(code, reason) which allows additional information to be provided to the clients.

Is there a real-world use case for this?

For context, in 12 years of developing this library, I have never come across anyone caring about close codes or reasons. Either you have a good reason and I'm interested, or let's keep the API simple and send a 1001 like the asyncio API does.

darkhaniop:
> Is there a real-world use case for this?

Admittedly, I have not used it in my client scripts either (as in, never made my scripts check the closing message). I added these args because the `ServerConnection.close()` accepts them, but I don't have strong opinions regarding their usefulness. So, I do not see an issue with not having them.

Speaking of similarity to asyncio, I see that the close method in `.asyncio.server.Server` accepts an optional `close_connections: bool = True` arg. Would adding the same option to `.sync.Server.shutdown()` be helpful?

aaugustin:
A good implementation is going to be substantially larger. Some locks are necessary; you don't have any; that doesn't reassure me that you're on top of all issues that could result from race conditions. I'm going to restart from scratch.

aaugustin:
Cf. https://github.com/python-websockets/websockets/pull/1736

PR #1632, opened 2025-06-07
More explicit error message when redirected to a non websocket uri

anthony-moreau:
When using the asyncio client, the error message and exception raised is the same if the user provides an incorrect uri or if the server respondes with a 302 found with a location header pointing to an incorrect uri.

This is quite confusing as it took me some time and debugging of the websocket library to figure out that the problem wasn't directly in my code but rather in the response of the server.

The current behavior isn't a bug as the uri was indeed a webpage and not a websocket so an exception in this scenario is normal. However a more explicit error message would help users understand the exact issue better.

As a result I'm suggesting changes to catch the InvalidUri exception in this scenario and modify the error message to make it more explicit. I also added a test to reflect theses changes.

### DIFF
--- a/src/websockets/asyncio/client.py
+++ b/src/websockets/asyncio/client.py
@@ -18,6 +18,7 @@ from ..exceptions import (
     InvalidProxyMessage,
     InvalidProxyStatus,
     InvalidStatus,
+    InvalidURI,
     ProxyError,
     SecurityError,
 )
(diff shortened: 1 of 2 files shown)


aaugustin:
Back when you submitted it, I noticed that the error message was getting heavy and repetitive. However, I didn't have an obvious idea to do better.

Options include:

- introducing an InvalidRedirectURI exception type;
- refactoring InvalidURI to customize the message — must be backwards compatible;
- other?

I haven't worked on this project in the past few weeks. I'm likely to have more time during the summer.

anthony-moreau:
I took some time to look again at my changes and you're right. I didn't take the time to read the definition of the InvalidUri exception so what I did wasn't optimal.

I decided to add a new InvalidRedirectURI exception subclassing InvalidUri with a different message as it looks to me like the simplest solution to the problem.

Feel free to take a look at it when you have the time to do so.

PR #1639, opened 2025-07-14
Fix `from websockets.(asyncio|sync).router import *` without werkzeug

hroncok:
This is a followup for 3128f5619de15991ec2af93d68c9c6626c6a5ae0

Also, skip test_router_supports_unix_sockets without werkzeug.

### DIFF
--- a/src/websockets/asyncio/router.py
+++ b/src/websockets/asyncio/router.py
@@ -35,6 +35,10 @@ except ImportError:
     ) -> Awaitable[Server]:
         raise ImportError("unix_route() requires werkzeug")
 
+    class Router:
+        def __init__(self, *args: Any, **kwargs: Any) -> None:
+            raise ImportError("Router() requires werkzeug")
+
 else:
 
     class Router:
--- a/src/websockets/sync/router.py
+++ b/src/websockets/sync/router.py
@@ -35,6 +35,10 @@ except ImportError:
     ) -> Server:
         raise ImportError("unix_route() requires werkzeug")
 
+    class Router:
+        def __init__(self, *args: Any, **kwargs: Any) -> None:
+            raise ImportError("Router() requires werkzeug")
+
 else:
 
     class Router:
--- a/tests/asyncio/test_router.py
+++ b/tests/asyncio/test_router.py
@@ -186,6 +186,7 @@ class RouterTests(EvalShellMixin, unittest.IsolatedAsyncioTestCase):
 
 
 @unittest.skipUnless(hasattr(socket, "AF_UNIX"), "this test requires Unix sockets")
+@unittest.skipUnless("werkzeug" in sys.modules, "werkzeug not installed")
 class UnixRouterTests(EvalShellMixin, unittest.IsolatedAsyncioTestCase):
     async def test_router_supports_unix_sockets(self):
         """Router supports Unix sockets."""
--- a/tests/sync/test_router.py
+++ b/tests/sync/test_router.py
@@ -162,6 +162,7 @@ class RouterTests(EvalShellMixin, unittest.TestCase):
 
 
 @unittest.skipUnless(hasattr(socket, "AF_UNIX"), "this test requires Unix sockets")
+@unittest.skipUnless("werkzeug" in sys.modules, "werkzeug not installed")
 class UnixRouterTests(EvalShellMixin, unittest.IsolatedAsyncioTestCase):
     def test_router_supports_unix_sockets(self):
         """Router supports Unix sockets."""


aaugustin:
The failed test run above is unrelated to your changes; it's backwards-incompatible changes in mypy :-(

aaugustin:
mypy doesn't like the override :-(

I see at least two options:

1. Move the `Router` class outside of the `try / except ImportError` block. Indeed, defining the class doesn't depend on werkzeug, only running it.
2. `# type: ignore`.

hroncok:
Which one do you prefer?

aaugustin:
Skipping tests that depend on werkzeug when werkzeug isn't installed is clearly correct.

I'm less clear about what needs to be done for the `Router` class; after further investigation, I believe the problem I must solve is this one:

```
>>> from websockets.asyncio.router import *
  File "<python-input-0>", line 1, in <module>
    from websockets.asyncio.router import *
AttributeError: module 'websockets.asyncio.router' has no attribute 'Router'. Did you mean: 'route'?
```

Note that it would not solve the problem described in https://github.com/fedora-eln/eln/issues/276 because the dependency on werkzeug was [explicitly declared](https://src.fedoraproject.org/rpms/python-websockets/blob/de584a09b8b71504daf5ea565b33c5244e2bc41f/f/python-websockets.spec#_20-21) and it will still be there regardless of the changes we're discussing here.

----

While we're discussing packaging, if Fedora wants to package websockets officially, it would be good to follow these guidelines e.g. to make sure that the code you're packaging is actually the code that I'm releasing: https://websockets.readthedocs.io/en/stable/project/contributing.html#packaging 

I am also puzzled by the decision to run a random subset of tests — why run the asyncio tests but not the sync tests? — and to skip failing tests — this defeats the point of running tests in the first place, doesn't it? It's burning CPU and creating manual work to add arbitrary tests to the list of failing tests when a build server gets a load spike while running the tests and a test fails randomly; I don't see how that creates value.

hroncok:
> I'm less clear about what needs to be done for the `Router` class; after further investigation, I believe the problem I must solve is this one:
> 
> ```
> >>> from websockets.asyncio.router import *
> Traceback (most recent call last):
>   File "<python-input-0>", line 1, in <module>
>     from websockets.asyncio.router import *
> AttributeError: module 'websockets.asyncio.router' has no attribute 'Router'. Did you mean: 'route'?
> ```

My change fixes the issue but breaks the typing check. Always defining the class but raising from `__init__` when workzeug is not importable is also a solution.

---

> Note that it would not solve the problem described in [fedora-eln/eln#276](https://github.com/fedora-eln/eln/issues/276) because the dependency on werkzeug was [explicitly declared](https://src.fedoraproject.org/rpms/python-websockets/blob/de584a09b8b71504daf5ea565b33c5244e2bc41f/f/python-websockets.spec#_20-21) and it will still be there regardless of the changes we're discussing here.

That's why I proposed to remove that declaration together with this change, see https://src.fedoraproject.org/rpms/python-websockets/pull-request/12 (churchyard is my username in Fedora).

> While we're discussing packaging, if Fedora wants to package websockets officially, it would be good to follow these guidelines e.g. to make sure that the code you're packaging is actually the code that I'm releasing: https://websockets.readthedocs.io/en/stable/project/contributing.html#packaging

I am here as an innocent drive-by contributor; the Fedora websockets package is not packaged by me. I merely wanted to drop the dependency declared by the Fedora maintainer.

> I am also puzzled by the decision to run a random subset of tests — why run the asyncio tests but not the sync tests? — and to skip failing tests — this defeats the point of running tests in the first place, doesn't it?

I agree, and I also proposed to run them all: https://src.fedoraproject.org/rpms/python-websockets/pull-request/14

Also, run them on all architectures: https://src.fedoraproject.org/rpms/python-websockets/pull-request/13

hroncok:
What sort of verification would we run instead? E.g. when we [update to a new Python version](https://fedoraproject.org/wiki/Changes/Python3.14) before you add it to your CI, how do we verify websocets isn't entirely broken?

aaugustin:
Fair point. I don't have a better answer to this than "run the tests". This allows you to validate automatically.

aaugustin:
In the case of Python 3.14, interestingly, the code is fine but the tests need a fix to pass without warnings :-)

hroncok:
Indeed, there is:

```
tests/legacy/utils.py:28: 526 warnings
  /builddir/build/BUILD/python-websockets-15.0.1-build/websockets-15.0.1/tests/legacy/utils.py:28: DeprecationWarning: 'asyncio.iscoroutinefunction' is deprecated and slated for removal in Python 3.16; use inspect.iscoroutinefunction() instead
    if asyncio.iscoroutinefunction(test):
```

hroncok:
Thanks.

PR #1660, opened 2025-09-19
allow user pass  ping_data in connection

liuyangc3:


aaugustin:
You didn't explain why you're proposing this, making it hard for me to tell if it's the right solution.

Technically, your code will break if `ping_timeout is None` and you don't receive a pong within `ping_interval`.

aaugustin:
The next step would be for me to understand what is the purpose of this feature and to decide whether I want to support it in the library.

aaugustin:
You may be proposing this because some servers require clients to send a ping with a specific content at a minimum frequency. However, most of those servers expect it in a Text frame, not a Ping frame. With the proposed `ping_data` API, users are likely to get confused.

Servers relying on protocol-level ping/pong in Ping/Pong frames, as opposed to application-level ping/pong in Text frames, usually don't care about the content of pings and pongs. websockets has been sending random 4-bytes payloads in pings for a decade without issues.

Issue #1675, opened 2025-11-18
Allow CLI to connect to custom "insecure" wss endpoints


abex-py:
When testing using the CLI tool, doing things quick'n'dirty for testing purpose is, imho, acceptable, as long as it's explicitly stated in the CLI args, for example something like this?
```bash
python -m websockets --insecure wss://localhost:8085
```

Imagine the following auto certif:
```bash
openssl req -x509 -newkey rsa:2048 -keyout key.pem -out cert.pem -days 365 -nodes -subj "/CN=localhost"
```

for this server:
```python
#!/usr/bin/env python

import asyncio
import ssl

from websockets.asyncio.server import serve


async def handler(websocket):
    while True:
        message = await websocket.recv()
        print(message)


async def main():
    ssl_context = ssl.SSLContext(ssl.PROTOCOL_TLS_SERVER)
    ssl_context.load_cert_chain(certfile="cert.pem", keyfile="key.pem")

    async with serve(handler, "localhost", 8085, ssl=ssl_context) as server:
        await asyncio.Future()  # run forever

if __name__ == "__main__":
    asyncio.run(main())
```

PR #1676, opened 2025-11-18
Add "--insecure" for wss endpoints #1675

abex-py:


aaugustin:
Fixed in b157ce05f728d4b6e326f32854c9033cd5992211. Same idea, with a slightly cleaner implementation, also with tests and docs.

PR #1683, opened 2025-12-11
Add --insecure flag to CLI for wss:// connections

### LINKED ISSUE #1675: Allow CLI to connect to custom "insecure" wss endpoints

majiayu000:
## Summary
- Add `--insecure` option to the CLI that allows connecting to wss:// endpoints with self-signed or invalid SSL certificates
- This is useful for testing and development purposes when using custom certificates

## Changes
- Modified `src/websockets/cli.py` to add `--insecure` argument
- When `--insecure` is specified, the CLI creates an SSL context that disables certificate verification and hostname checking
- Added tests for the new functionality

## Test plan
- [x] Added test `test_insecure_connection` - verifies connection works with `--insecure` flag
- [x] Added test `test_insecure_connection_fails_without_flag` - verifies connection fails without the flag when using self-signed cert
- [x] All existing tests pass

Fixes #1675

### DIFF
--- a/src/websockets/cli.py
+++ b/src/websockets/cli.py
@@ -3,8 +3,9 @@ from __future__ import annotations
 import argparse
 import asyncio
 import os
+import ssl
 import sys
-from typing import Generator
+from typing import Any, Generator
 
 from .asyncio.client import ClientConnection, connect
 from .asyncio.messages import SimpleQueue
@@ -101,9 +102,9 @@ async def send_outgoing_messages(
             break
 
 
-async def interactive_client(uri: str) -> None:
+async def interactive_client(uri: str, **kwargs: Any) -> None:
     try:
-        websocket = await connect(uri)
+        websocket = await connect(uri, **kwargs)
     except Exception as exc:
         print(f"Failed to connect to {uri}: {exc}.")
         sys.exit(1)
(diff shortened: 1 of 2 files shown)


aaugustin:
Your PR ignores my comment [here](https://github.com/python-websockets/websockets/issues/1675#issuecomment-3557159790) — AI slop?

PR #1685, opened 2025-12-11
Update memory docs for threading max_queue parameter

### LINKED ISSUE #1664: "Memory and Buffers" docs and max_queue in threading implementation
I believe the [topic page "Memory and Buffers" in the documentation](https://websockets.readthedocs.io/en/stable/topics/memory.html#websockets-buffers) might need to be updated - in the end it says:

> In the [threading](https://docs.python.org/3/library/threading.html#module-threading) implementation, there is no queue of incoming messages. The max_queue argument doesn’t exist. The connection keeps at most one message in memory at a time.

But looks like max_queue exists in the threading implementation - I guess works the same as in the asyncio one?

majiayu000:
## Summary
Update the Memory and Buffers documentation to reflect that the threading implementation now has a `max_queue` parameter with the same semantics as the asyncio implementation.

## Changes
- Updated `docs/topics/memory.rst` to document that the threading implementation has `max_queue`

Fixes #1664

### DIFF
--- a/docs/topics/memory.rst
+++ b/docs/topics/memory.rst
@@ -152,6 +152,6 @@ and :func:`~asyncio.server.serve` varies across implementations.
   use up to 4 bytes of memory per character, the queue may use up to ``4 *
   max_size * max_queue`` bytes of memory. By default, this is 128 MiB.
 
-* In the :mod:`threading` implementation, there is no queue of incoming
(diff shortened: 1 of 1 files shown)


aaugustin:
Fixed (a bit differently) in 6563a9c884d92df4c889bb174dc84a56a7377686.

PR #1719, opened 2026-06-20
Add text argument to broadcast() to force the frame type

### LINKED ISSUE #1694: Add option to force sending text or binary frames to asyncio broadcast()
A `text` argument to `broadcast()` similar to how `send()` handles it would allow me to change a line like 

`websockets.broadcast(connections, data.decode("utf-8"))` (only to immediately encode it again) to
`websockets.broadcast(connections, data, text=True)`

which would also relieve the code at that point from knowing whether `data` is bytes or str, only that it needs to go into a "Text" frame.

I've been testing this change to asyncio-broadcast so far with no ill effects.

src/websockets/asyncio/connection.py:
```
def broadcast(
    connections: Iterable[Connection],
    message: DataLike,
    raise_exceptions: bool = False,
    *,
    text: bool | None = None,
) -> None:
"""
...

"""
    if isinstance(message, str):
        if text is None:
            send_method = "send_text"
        message = message.encode()
    elif isinstance(message, BytesLike):
        if text is None:
            send_method = "send_binary"
    else:
        raise TypeError("data must be str or bytes")

    if text is not None:
        send_method = "send_text" if text else "send_binary"
```

( in a [branch here](https://github.com/python-websockets/websockets/compare/main...lampeh:websockets:broadcast-text) )

What would be the chances of getting type-overrides backported to 13.x? (Mostly j/k, but I have one test environment with Python 3.8/websockets 13 left and the patch [applies the same](https://github.com/python-websockets/websockets/compare/13.x...lampeh:websockets:broadcast-text_13.x) to the 13.x branch but lacks the change to `send()`)

[broadcast-text.patch](https://github.com/user-attachments/files/24763180/broadcast-text.patch)

eugen-goebel:
### What this does

`broadcast()` now accepts a keyword-only `text` argument that mirrors `Connection.send()`:

* `text=True` sends a bytestring or bytes-like object in a Text frame.
* `text=False` sends a `str` in a Binary frame.
* `text=None` (default) keeps the current behavior, choosing the frame type from the message type.

This lets callers control the frame type without converting the message first. For example, an already UTF-8 encoded payload can be broadcast as a text frame without decoding and re-encoding it.

Closes #1694.

### Credit

The implementation mirrors the semantics of `Connection.send()`. Credit to @lampeh, who proposed this in #1694 and provided a patch.

### Tests and docs

* Added `test_broadcast_text_from_bytes` and `test_broadcast_binary_from_str` in `tests/asyncio/test_connection.py`, matching the existing `send()` override tests. All four `text` branches are exercised, so coverage stays at 100%.
* Documented the `text` argument in the `broadcast()` docstring, mirroring `send()`.
* Added a changelog entry under 16.1.
* `ruff format --check` and `ruff check` pass on `src` and `tests`.

### DIFF
--- a/docs/project/changelog.rst
+++ b/docs/project/changelog.rst
@@ -42,6 +42,10 @@ Improvements
 
 * Added wheels for ARMv7, PowerPC, RISC-V, and S/390.
 
+* Added the ``text`` argument to :func:`~asyncio.server.broadcast`, mirroring
+  :meth:`~asyncio.connection.Connection.send`, to force sending a Text or
+  Binary frame regardless of the type of ``message``.
+
 Bug fixes
 .........
 
--- a/src/websockets/asyncio/connection.py
+++ b/src/websockets/asyncio/connection.py
@@ -1148,6 +1148,8 @@ def broadcast(
     connections: Iterable[Connection],
     message: DataLike,
     raise_exceptions: bool = False,
+    *,
+    text: bool | None = None,
 ) -> None:
     """
     Broadcast a message to several WebSocket connections.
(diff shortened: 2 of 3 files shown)


aaugustin:
Looks good! Thank you for your contribution.

Issue #1592, opened 2025-02-09
`python -m websockets` blocks until open_timeout when connection fails


aaugustin:
```
$ python -m websockets wss://.../
^C
Traceback (most recent call last):
  File "/Users/myk/dev/websockets/src/websockets/sync/client.py", line 378, in connect
    connection.handshake(
    ~~~~~~~~~~~~~~~~~~~~^
        additional_headers,
        ^^^^^^^^^^^^^^^^^^^
        user_agent_header,
        ^^^^^^^^^^^^^^^^^^
        deadline.timeout(),
        ^^^^^^^^^^^^^^^^^^^
    )
    ^
  File "/Users/myk/dev/websockets/src/websockets/sync/client.py", line 101, in handshake
    raise self.protocol.handshake_exc
  File "/Users/myk/dev/websockets/src/websockets/client.py", line 325, in parse
    self.process_response(response)
    ~~~~~~~~~~~~~~~~~~~~~^^^^^^^^^^
  File "/Users/myk/dev/websockets/src/websockets/client.py", line 142, in process_response
    raise InvalidStatus(response)
websockets.exceptions.InvalidStatus: server rejected WebSocket connection: HTTP 302

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "<frozen runpy>", line 198, in _run_module_as_main
  File "<frozen runpy>", line 88, in _run_code
  File "/Users/myk/dev/websockets/src/websockets/__main__.py", line 159, in <module>
    main()
    ~~~~^^
  File "/Users/myk/dev/websockets/src/websockets/__main__.py", line 131, in main
    websocket = connect(args.uri)
  File "/Users/myk/dev/websockets/src/websockets/sync/client.py", line 385, in connect
    connection.recv_events_thread.join()
    ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~^^
  File "/Users/myk/.pyenv/versions/3.13.1/lib/python3.13/threading.py", line 1092, in join
    self._handle.join(timeout)
    ~~~~~~~~~~~~~~~~~^^^^^^^^^
KeyboardInterrupt
```

Connecting to the same server directly with `websockets.sync.client.connect` doesn't block.

aaugustin:
Actually this is a race condition.

When this line:

https://github.com/python-websockets/websockets/blob/8c9f6fc27bb54610f1c389d7769cddc1504ca57e/src/websockets/sync/connection.py#L814

runs before this line:

[...]

PR #1721, opened 2026-07-03
Stop reading when the opening handshake fails.

### LINKED ISSUE #1596: Race condition in sync implemention delays closing connections
When this line:

https://github.com/python-websockets/websockets/blob/8c9f6fc27bb54610f1c389d7769cddc1504ca57e/src/websockets/sync/connection.py#L814

runs before this line:

https://github.com/python-websockets/websockets/blob/8c9f6fc27bb54610f1c389d7769cddc1504ca57e/src/websockets/sync/client.py#L384

closing the socket doesn't interrupt `recv()` because `recv()` isn't running in the main thread — and interruption is based on signals which are only received by the main thread.

First discovered in #1592.

scttbnsn:
Fix #1596.

I could reproduce the delay on macOS (Python 3.13) when the server rejects the handshake and keeps the TCP connection open, for both `ws://` and `wss://`, at a 30-75% rate per attempt. The mechanism turned out a little different from the issue description.

After parsing the non-101 response, the protocol sends EOF, so `send_data()` half-closes the socket with `shutdown(SHUT_WR)` and `recv_events()` blocks in a `recv()` bounded by `close_deadline`, waiting for the server to close the connection. When `close_socket()` then runs in the thread that called `connect()`, its `shutdown(SHUT_RDWR)` is a second `shutdown()` on a half-closed socket. On macOS that raises `ENOTCONN`, which the existing `except OSError: pass` swallows, and `close()` doesn't reliably interrupt a `recv()` that's already blocked. `recv_events()` rides out `close_timeout`, or blocks forever with `close_timeout=None`. On Linux the second `shutdown()` succeeds, which is why this doesn't reproduce there.

About the issue description: the explanation based on signals didn't hold up when I tested it. Signals are indeed delivered only to the main thread, which is why Ctrl-C in #1592 interrupted `join()` rather than `recv()`, but waking a blocked `recv()` with `shutdown()` and `close()` doesn't involve signals. As long as the socket isn't already half-closed, it works reliably from any thread; I measured 20/20 prompt wakeups with `recv()` blocked in a background thread and 10/10 with the roles reversed, `recv()` in the main thread and the closing calls in a background thread. The ordering described in the issue is real: `recv_events()` has to be blocked in `recv()` before `close_socket()` runs. What makes the wakeup fail is the prior half-close, not which thread `recv()` runs on.

I couldn't find any combination of `shutdown()`/`close()` calls that interrupts the blocked `recv()` reliably on macOS, so the fix removes the wait instead: when the opening handshake failed, the connection is unusable and `close_socket()` runs as soon as the failure propagates, so there's no point waiting for the peer to close the connection. `recv_events()` stops reading and tears down the socket in its own thread, without depending on cross-thread wakeup. With this change I measured 0 delayed teardowns in 80 attempts, versus 30 in 40 before.

The other scenario, where the server never responds and `open_timeout` fires, didn't hang in my tests: there's no prior half-close, so the first `shutdown()` succeeds and interrupts the initial `recv()` reliably. It's unchanged.

The client test simulates the macOS `shutdown()`/`close()` semantics at the socket level, so it fails without the fix on any OS, including Linux CI. The connection-level test exercises the new branch for both sides, which also keeps `maxi_cov` at 100%.

### DIFF
--- a/docs/project/changelog.rst
+++ b/docs/project/changelog.rst
@@ -117,6 +117,9 @@ Bug fixes
 
 * Restored compatibility of the ``websockets`` CLI with Windows.
 
+* Fixed a race condition that could delay closing connections in the
+  :mod:`threading` implementation.
+
 .. _16.1.1:
 
 16.1.1
--- a/src/websockets/sync/connection.py
+++ b/src/websockets/sync/connection.py
@@ -22,7 +22,7 @@ from ..exceptions import (
 )
 from ..frames import DATA_OPCODES, CloseCode, Frame, Opcode
 from ..http11 import Request, Response
-from ..protocol import CLOSED, OPEN, Event, Protocol, State
+from ..protocol import CLOSED, CONNECTING, OPEN, Event, Protocol, State
 from ..typing import BytesLike, Data, DataLike, LoggerLike, Subprotocol
 from .messages import Assembler
 from .utils import Deadline
@@ -870,6 +870,14 @@ class Connection:
                         if self.close_deadline is None:
                             self.close_deadline = Deadline(self.close_timeout)
 
+                    # If the opening handshake failed, the connection never
+                    # reached the OPEN state and there's no closing handshake
+                    # to wait for. Remember this so we can stop reading below.
+                    handshake_failed = (
+                        self.protocol.state is CONNECTING
+                        and self.protocol.close_expected()
+                    )
+
                 # Unlock conn_mutex before processing events. Else, the
                 # application can't send messages in response to events.
 
@@ -880,6 +888,17 @@ class Connection:
                 for event in events:
                     # This isn't expected to raise an exception.
                     self.process_event(event)
+
+                # When the opening handshake fails, the connection is unusable
+                # and close_socket() runs as soon as the handshake failure
+                # propagates. Stop reading instead of waiting for the peer to
+                # close the TCP connection, because closing a socket in one
+                # thread doesn't always interrupt recv() in another thread --
+                # e.g. on macOS, when the protocol already half-closed the
+                # socket, shutdown() fails with ENOTCONN and close() doesn't
+                # interrupt recv().
+                if handshake_failed:
+                    break
 
             # Breaking out of the while True: ... loop means that we believe
             # that the socket doesn't work anymore.
--- a/tests/sync/test_client.py
+++ b/tests/sync/test_client.py
@@ -1,3 +1,4 @@
+import errno
 import http
 import logging
 import os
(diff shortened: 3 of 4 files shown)


scttbnsn:
I also traced the server-side rejection path for the same problem. When `process_request` rejects the handshake, `send_context()` half-closes the socket from the handler thread, and its forced `close_socket()` after `close_timeout` is again a second `shutdown()`, so the swallowed ENOTCONN happens there too. I couldn't turn it into a hang though: `recv_events()` is blocked in an untimed `recv()` on that path and `close()` woke it in 30/30 attempts on macOS, so this PR leaves it alone.

aaugustin:
~As you must have noticed, you're touching one of the trickiest areas of websockets' codebase! It's relatively recent and I'm willing to believe that not all error paths are handled correctly.~

I'll look carefully into your analysis and proposal.

~If you believe some of the content in the original issue is hallucinated (it's clearly AI-generated), please edit the description to make that clear 🙏~

EDIT: I was confused when made this comment; I thought it was for #1722 which you created around the same time.

scttbnsn:
I edited the description to spell out what I think doesn't hold up in the issue: the signal-based explanation. Signal delivery being main-thread-only is real and explains the traceback in #1592, but it's not why closing the socket fails to interrupt `recv()`. `shutdown()` + `close()` wakes a blocked `recv()` from any thread, unless the socket was already half-closed by the EOF the protocol sends after parsing the rejection. That prior half-close is the ingredient the issue was missing.

One nuance I found while probing: even though the second `shutdown()` fails with `ENOTCONN`, attempting it still makes the subsequent `close()` much more likely to wake the blocked `recv()` on macOS (93% vs 0% in isolated probes when skipping it entirely). So the swallowed call in `close_socket()` is doing useful work and I left it alone.

I also fixed the CI failure: `maxi_cov` measures `sync/connection.py` against `test_connection.py` alone, and the new branch was only exercised from `test_client.py`. I added a connection-level test that simulates a failed opening handshake the way `ClientProtocol.parse` leaves the protocol, runs for both sides, and fails without the fix by riding out `close_timeout`. Amended into the same commit.

aaugustin:
> there's no point waiting for the peer to close the connection

In theory, the WebSocket RFC says:

> As such, when a server
   is instructed to _Close the WebSocket Connection_ it SHOULD initiate
   a TCP Close immediately, and when a client is instructed to do the
   same, it SHOULD wait for a TCP Close from the server.

> As an example of how to obtain a clean closure in C using Berkeley
   sockets, one would call shutdown() with SHUT_WR on the socket, call
   recv() until obtaining a return value of 0 indicating that the peer
   has also performed an orderly shutdown, and finally call close() on
   the socket.

In practice, when we send a non 101-response, we can argue that it's HTTP not WebSocket and the HTTP RFC has no such constraints.

aaugustin:
Will finish this in #1738.

Issue #1716, opened 2026-06-12
Tests fail on single-cpu machines (possibly due to a race condition)


sanvila:
Hello.

When I build the Debian package for this project on AWS instances of type `m7a.medium` or `r7a.medium`, which incidentally have a single CPU, the following two tests usually fail:

```
FAIL: test_writing_in_send_context_fails (tests.sync.test_connection.ClientConnectionTests.test_writing_in_send_context_fails)
[...]
FAIL: test_writing_in_send_context_fails (tests.sync.test_connection.ServerConnectionTests.test_writing_in_send_context_fails)
```

Sometimes, they give an error instead:

```
ERROR: test_writing_in_recv_events_fails (tests.sync.test_connection.ClientConnectionTests.test_writing_in_recv_events_fails)
[...]
ERROR: test_writing_in_recv_events_fails (tests.sync.test_connection.ServerConnectionTests.test_writing_in_recv_events_fails)
```

And there are also times where one test fails and the other gives an error.

This is an example of failure:

```
======================================================================
FAIL: test_writing_in_send_context_fails (tests.sync.test_connection.ClientConnectionTests.test_writing_in_send_context_fails)
Error when sending outgoing frame is correctly reported.
----------------------------------------------------------------------
Traceback (most recent call last):
  File "/<<PKGBUILDDIR>>/.pybuild/cpython3_3.13_websockets/build/tests/sync/test_connection.py", line 974, in test_writing_in_send_context_fails
    self.assertIsInstance(raised.exception.__cause__, BrokenPipeError)
    ~~~~~~~~~~~~~~~~~~~~~^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
AssertionError: None is not an instance of <class 'BrokenPipeError'>

======================================================================
FAIL: test_writing_in_send_context_fails (tests.sync.test_connection.ServerConnectionTests.test_writing_in_send_context_fails)
Error when sending outgoing frame is correctly reported.
----------------------------------------------------------------------
Traceback (most recent call last):
  File "/<<PKGBUILDDIR>>/.pybuild/cpython3_3.13_websockets/build/tests/sync/test_connection.py", line 974, in test_writing_in_send_context_fails
    self.assertIsInstance(raised.exception.__cause__, BrokenPipeError)
[...]

Issue #866, opened 2020-12-11
PayloadTooBig, frame size, and maximum payload question?


mortoray:
The Python client has a maximum payload size. I know I can increase this, but I had a question in general. Is this only the payload size? If  understand correctly though the server has no limit. It will fragment messages into frames. What is that size (I want to make sure I'm under the AWS size).  

Is there a real limit to the maximum message size in WebSocket itself, or is it only the Python client that does this? I'm only trying to understand to avoid designing something that won't work everywhere.

aaugustin:
This library doesn't automatically fragment messages. However, it allows you to fragment them as you wish by passing an iterator of fragments to `send()` instead of a single message.

The limit on the size of a message is only for incoming messages It exists for security reasons. Without the limit, you could trivially crash the server by sending a message of an impractically large size (likely in the form of a zip bomb, given that compression is enabled by default). 

There's a protocol-level limit on the size of a frame in the WebSocket protocol, which (if memory serves) is 2**63 - 1 bytes — compressed ;-)

This doesn't answer exactly your question but hopefull will point you in the right direction.

aaugustin:
BTW - thanks for taking the time to write https://mortoray.com/2020/12/06/high-throughput-game-message-server-with-python-websockets/ - it's always interesting to see how this library is used!

I share your opinion on acceptable throughput. Python is slow, sure, but 10k msg/sec in a single thread is table-stakes :-) I believe it's possible to do even better but I'm not sure it's _generically_ possible to do better, without knowing what OS you'll be running on.

mortoray:
Thank you. I'll need to do my own message fragmentation then. I'll test that, and if it works I'll close this issue.

The incoming limit makes sense.

Your welcome, about the article. I'll likely write a followup with the higher-level pieces.  10k/s supports a 0.1ms ping-time... which would be a response-time dream on the internet!  :D

mortoray:
I've tried fragmented, by passing a list of bytes to the `send` method. The Python client accepts this fine, but a browser WebSocket connection rejects it -- the connection is closed.

My chunking code is this:

```
def encode_msg(msg: Dict) -> List[bytes]:
	all_bytes = json.dumps(msg, ensure_ascii=False).encode('utf-8')
	chunks = (len(all_bytes)-1) // fragment_size + 1
	if chunks > 1:
		print("Chunking",chunks)
	return [all_bytes[x*fragment_size:(x+1)*fragment_size] for x in range(0,chunks)]
```

[...]

PR #1723, opened 2026-07-03
Reject permessage-deflate continuation frames with RSV1 set

### LINKED ISSUE #1720: permessage-deflate accepts continuation frames with RSV1 set
## Summary

`websockets==16.0` appears to accept a `permessage-deflate` continuation frame when the RSV1 bit is set.

My reading of RFC 7692 is that RSV1 marks the first frame of a compressed message. It should not be set on continuation frames. If a continuation frame has RSV1 set, the connection should fail with a protocol error.

## Reproduction

This small test shows the behavior at the extension layer:

```python
import zlib

import pytest

from websockets.exceptions import ProtocolError
from websockets.extensions.permessage_deflate import PerMessageDeflate
from websockets.frames import Frame, OP_CONT, OP_TEXT


def compressed_message(data: bytes) -> bytes:
    compressor = zlib.compressobj(wbits=-15)
    compressed = compressor.compress(data) + compressor.flush(zlib.Z_SYNC_FLUSH)
    return compressed[:-4]


def test_permessage_deflate_rejects_rsv1_on_continuation():
    extension = PerMessageDeflate(False, False, 15, 15)

    compressed = compressed_message(b'{"cmd":"admin"}')
    split = max(1, len(compressed) // 2)

    extension.decode(Frame(OP_TEXT, compressed[:split], fin=False, rsv1=True))

    with pytest.raises(ProtocolError):
        extension.decode(Frame(OP_CONT, compressed[split:], fin=True, rsv1=True))
```

## Expected behavior

The second frame should raise `ProtocolError` because it is a continuation frame with `RSV1=1`.

## Actual behavior

The second frame is accepted and decompressed.

I also reproduced this end to end with `permessage-deflate` negotiated:

1. Send a first text frame with `FIN=0` and `RSV1=1`.
2. Send a final continuation frame with `FIN=1` and `RSV1=1`.

The message is delivered to the application as:

```text
{"cmd":"admin"}
```

## Possible cause

In `PerMessageDeflate.decode()`, the continuation-frame branch checks whether continuation data should be decoded, but it does not appear to reject `frame.rsv1`. The returned frame then has `rsv1=False`, so later frame validation no longer sees the RSV1 bit from the original continuation frame.

## Possible fix

One possible fix is to reject RSV1 before decoding continuation-frame extension data:

```diff
diff --git a/src/websockets/extensions/permessage_deflate.py b/src/websockets/extensions/permessage_deflate.py
@@
         if frame.opcode is frames.OP_CONT:
+            if frame.rsv1:
+                raise ProtocolError("reserved bits must be 0")
             if not self.decode_cont_data:
                 return frame
             if frame.fin:
                 self.decode_cont_data = False
```

## References

RFC 7692, section 6.2:
https://datatracker.ietf.org/doc/html/rfc7692#section-6.2

RFC 7692, section 7.2.2:
https://datatracker.ietf.org/doc/html/rfc7692#section-7.2.2

Sanjays2402:
Fixes #1720.

## Root cause

Per [RFC 7692, section 6.2](https://datatracker.ietf.org/doc/html/rfc7692#section-6.2), the RSV1 bit marks the **first** frame of a compressed message. It must not be set on continuation frames; a continuation frame with RSV1 set is a protocol error.

`PerMessageDeflate.decode()` handled the RSV1 bit only in the first-frame branch. The continuation branch (`frame.opcode is frames.OP_CONT`) never inspected `frame.rsv1`, and the method always returns the frame with `rsv1` unset:

```python
return frames.Frame(
    frame.opcode,
    data,
    frame.fin,
    False,  # rsv1 always cleared
    frame.rsv2,
    frame.rsv3,
)
```

So an illegal RSV1 bit on a **compressed** continuation frame was silently swallowed before `Frame.check()` (which raises `ProtocolError("reserved bits must be 0")`) could ever see it. As @aaugustin noted in the issue, the symmetric send side was already fixed in #866; the receive side was missed.

## Before

```python
extension = PerMessageDeflate(False, False, 15, 15)
compressed = compressed_message(b'{"cmd":"admin"}')
split = len(compressed) // 2

extension.decode(Frame(OP_TEXT, compressed[:split], fin=False, rsv1=True))   # legal first frame
extension.decode(Frame(OP_CONT, compressed[split:], fin=True, rsv1=True))    # illegal: accepted and decoded
```

| Frame | Before this PR | After this PR |
|-------|---------------|---------------|
| `OP_CONT`, `rsv1=1`, compressed | accepted, decompressed | `ProtocolError: reserved bits must be 0` |
| `OP_CONT`, `rsv1=1`, uncompressed | passed through with `rsv1=1` (caught later by `Frame.check()`) | `ProtocolError: reserved bits must be 0` (rejected at the extension layer, consistently) |

## Fix

Reject the RSV1 bit at the top of the continuation branch, reusing the existing `"reserved bits must be 0"` protocol error used by `Frame.check()`. Both compressed and uncompressed continuation frames are now rejected consistently at the extension layer.

```diff
         # Handle continuation data frames:
+        # - reject the rsv1 flag, which is only valid on the first frame
         # - skip if the message isn't encoded
         # - reset "decode continuation data" flag if it's a final frame
         if frame.opcode is frames.OP_CONT:
+            if frame.rsv1:
+                raise ProtocolError("reserved bits must be 0")
             if not self.decode_cont_data:
                 return frame
             if frame.fin:
                 self.decode_cont_data = False
```

## Tests

Two regression tests added to `tests/extensions/test_permessage_deflate.py` (compressed and uncompressed continuation frames). Proven to guard the bug by stashing only the source change:

- **Without the fix:** both tests `FAIL` with `AssertionError: ProtocolError not raised`.
- **With the fix:** both pass.

`test_permessage_deflate` is fully green (35 tests), as are the frames/protocol/extensions suites (288 tests). `ruff check`, `ruff format --check`, and `mypy` are clean on the changed files. The only failures in the full suite are pre-existing timing flakes in the `sync` tests (see #1716) and missing-`trio` import errors, both present on a clean baseline and unrelated to this change.

### DIFF
--- a/docs/project/changelog.rst
+++ b/docs/project/changelog.rst
@@ -48,6 +48,9 @@ Bug fixes
 * Prevented ``Frame.__str__`` from crashing when a text frame is fragmented in
   the middle of a UTF-8 sequence.
 
+* Rejected incoming ``permessage-deflate`` continuation frames with the RSV1
+  bit set, as required by :rfc:`7692`.
+
 .. _16.0:
 
 16.0
--- a/src/websockets/extensions/permessage_deflate.py
+++ b/src/websockets/extensions/permessage_deflate.py
@@ -104,9 +104,12 @@ class PerMessageDeflate(Extension):
             return frame
 
         # Handle continuation data frames:
+        # - reject the rsv1 flag, which is only valid on the first frame
         # - skip if the message isn't encoded
         # - reset "decode continuation data" flag if it's a final frame
         if frame.opcode is frames.OP_CONT:
+            if frame.rsv1:
+                raise ProtocolError("reserved bits must be 0")
             if not self.decode_cont_data:
                 return frame
             if frame.fin:
--- a/tests/extensions/test_permessage_deflate.py
+++ b/tests/extensions/test_permessage_deflate.py
@@ -213,6 +213,31 @@ class PerMessageDeflateTests(unittest.TestCase, PerMessageDeflateTestsMixin):
         self.assertEqual(dec_frame1, frame1)
         self.assertEqual(dec_frame2, frame2)
 
+    def test_decode_rsv1_on_compressed_continuation_frame(self):
+        # RSV1 marks the first frame of a compressed message; it must not be
+        # set on continuation frames (RFC 7692, section 6.2).
+        frame1 = Frame(OP_TEXT, "café".encode(), fin=False)
+        frame2 = Frame(OP_CONT, " & ".encode(), fin=False)
+
+        enc_frame1 = self.extension.encode(frame1)
+        enc_frame2 = self.extension.encode(frame2)
+
+        self.extension.decode(enc_frame1)
+
+        with self.assertRaises(ProtocolError):
+            self.extension.decode(dataclasses.replace(enc_frame2, rsv1=True))
+
+    def test_decode_rsv1_on_uncompressed_continuation_frame(self):
+        # An uncompressed message has no RSV1 on any of its frames, including
+        # continuation frames (RFC 7692, section 6.2).
+        frame1 = Frame(OP_TEXT, "café".encode(), fin=False)
+        frame2 = Frame(OP_CONT, " & ".encode(), fin=False)
+
+        self.extension.decode(frame1)
+
+        with self.assertRaises(ProtocolError):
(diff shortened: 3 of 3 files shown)


aaugustin:
Just realized I fixed it two days ago and forgot to push the commit (f729dc4), sorry.

PR #1724, opened 2026-07-03
Remove unused imports in docs/conf.py (F401)

Mubashir78:
Remove unused imports in the docs configuration flagged by Ruff rule F401.

### DIFF
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -24,7 +24,7 @@ project = "websockets"
 copyright = f"2013-{datetime.date.today().year}, Aymeric Augustin and contributors"
 author = "Aymeric Augustin"
 
-
+from websockets.version import tag as version, version as release
 
 # -- General configuration ---------------------------------------------------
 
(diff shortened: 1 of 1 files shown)


aaugustin:
Did you confirm that the docs build correctly after making those changes?

Removing subprocess in correct; I'm pretty sure removing version / tag is wrong.

Mubashir78:
Yeah you're right, `version` and `release` are Sphinx config variables read from the conf.py namespace at build time. I've reverted that removal while keeping `import subprocess` removed.

I confirmed the docs build succeeds with `sphinx-build -b html docs docs/_build/html` after the fix.

Hope all is good now!

aaugustin:
I don't want to just fix it once; instead I want to use the same tool that prevent stray imports elsewhere.

I did it in 99431eea503c70b57412fbeda43481d3081d6037.

aaugustin:
Thank you for surfacing the issue!

PR #1726, opened 2026-07-09
Handle rejected handshakes cleanly in the threading server

apoorvdarshan:
## Summary

The threading server didn't handle a rejected opening handshake explicitly. When `process_request` or `process_response` returned an HTTP response, `handshake()` returned normally while the protocol state wasn't `OPEN`. The following assertion failed silently—or disappeared under `python -O`, allowing the handler to run on a closed connection and log `"connection handler failed"`.

## Fix

Replace the assertion with an explicit rejection path that closes the socket, joins the receive thread, and returns, mirroring the asyncio implementation.

## Tests / Verification

- Guarded `test_process_request_returns_response` with `assertNoLogs("websockets", logging.ERROR)`.
- Matched the asyncio tests by verifying the exact `"opening handshake failed"` error and exception for `process_request` and `process_response` failures.
- Official coverage environment: **2,258 tests passed**, 10 skipped, **100% coverage**.
- Ruff formatting and lint checks pass.
- Strict mypy checks pass.
- `python -O -m unittest tests.sync.test_server`: **50 tests passed**.

Refs #1513.

Disclosure: prepared with AI assistance; manually reviewed and verified locally.

PR #1729, opened 2026-07-12
Fix rejected handshake cleanup in threading server

### LINKED ISSUE #1722: Rejecting a handshake raises a silenced AssertionError in the threading server
In the threading implementation, when `process_request` or `process_response` rejects the handshake by returning an HTTP response, `handshake()` returns normally; `handshake_exc` is only set for failures, not for deliberate rejections:

https://github.com/python-websockets/websockets/blob/ff4869ba468129f3e85b08c2a8a03ec45cf26537/src/websockets/sync/server.py#L180-L189

`conn_handler` then hits this assertion, since the state is CONNECTING or CLOSED after a rejection, never OPEN:

https://github.com/python-websockets/websockets/blob/ff4869ba468129f3e85b08c2a8a03ec45cf26537/src/websockets/sync/server.py#L608

The AssertionError is swallowed by the `except Exception` below, so nothing is logged and the socket is closed without going through `close_socket()` and joining `recv_events_thread`. The `# pragma: no cover` there isn't accurate either; this branch runs on every rejection:

https://github.com/python-websockets/websockets/blob/ff4869ba468129f3e85b08c2a8a03ec45cf26537/src/websockets/sync/server.py#L618-L620

With `python -O`, the assertion is stripped instead and the handler runs on the rejected connection: `start_keepalive()` and `handler(connection)` execute with the connection already CLOSED, and every rejection logs "connection handler failed".

Repro:

```python
import http
import threading

from websockets.sync.client import connect
from websockets.sync.server import serve


def reject(connection, request):
    return connection.respond(http.HTTPStatus.FORBIDDEN, "Forbidden")


def handler(connection):
    raise AssertionError("handler ran on a rejected connection")


with serve(handler, "localhost", 0, process_request=reject) as server:
    threading.Thread(target=server.serve_forever, daemon=True).start()
    host, port = server.socket.getsockname()
    try:
        connect(f"ws://{host}:{port}/")
    except Exception as exc:
        print(f"client: {exc!r}")
    server.shutdown()
```

Under `python`, the client gets the 403 and the server logs nothing; a trace hook shows the AssertionError firing at line 608 on every rejection. Under `python -O`, the server additionally logs "connection handler failed" with the handler's traceback.

Verified on main (ff4869b) with Python 3.13 on macOS.

Sanjays2402:
The threading server treated an intentional HTTP rejection as a successful handshake and relied on an assertion to stop before the connection handler. That assertion was swallowed in normal mode and removed under `python -O`, where the handler ran on a closed connection.

This returns explicitly for a rejected handshake, closes and joins the receive thread, and adds a regression check that detects the internal assertion. Fixes #1722.

### DIFF
--- a/docs/project/changelog.rst
+++ b/docs/project/changelog.rst
@@ -59,6 +59,12 @@ New features
 * Added the ``--insecure`` option to the ``websockets`` CLI to disable TLS
   certificate validation.
 
+Bug fixes
+.........
+
+* Avoided an internal assertion when the :mod:`threading` server rejects an
+  opening handshake. (:issue:`1722`)
+
 .. _16.1:
 
 16.1
--- a/src/websockets/sync/server.py
+++ b/src/websockets/sync/server.py
@@ -615,7 +615,10 @@ def serve(
                 connection.recv_events_thread.join()
                 return
 
-            assert connection.protocol.state is OPEN
+            if connection.protocol.state is not OPEN:
+                connection.close_socket()
+                connection.recv_events_thread.join()
+                return
             try:
                 connection.start_keepalive()
                 handler(connection)
--- a/tests/sync/test_server.py
+++ b/tests/sync/test_server.py
@@ -3,6 +3,7 @@ import hmac
 import http
 import logging
 import socket
+import threading
 import time
 import unittest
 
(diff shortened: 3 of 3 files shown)


aaugustin:
Duplicate of #1726.

aaugustin:
I prefer the approach to testing from #1726; it's more straightforward and consistent with pre-existing tests.

Issue #1059, opened 2021-09-27
Reference cycle in websockets 10?


chrysn:
In the CI tests of aiocoap, I noticed breakage going from websockets 9.1 to 10 without finding a plausible link to the change and commit log.

The change is subtle and usually not covered by Python's API guarantees: I check that after an asynchronous shutdown (which in the websockets case performs a `.close()` and `await s.wait_closed()` on the server), the large library objects are actually freed when their last reference is removed. This is useful in spotting memory leakage -- though granted, it may be overzealous.

Does the websockets package generally aim for the same property of running the destructors by refcount? If not, would you accept changes that make it so?

I do not have a minimal test case yet, hoping for a "yeah we changed X, you'll have to shut down more tightly by doing Y"; otherwise, I can try to assemble a test. More for my own reference and the very very curious: Public logs of the diverging behavior for [10](https://gitlab.com/aiocoap/aiocoap/-/jobs/1623608185) and [9.1](https://gitlab.com/aiocoap/aiocoap/-/jobs/1623695631), a minimal check in aiocoap's tests is conducted using `AIOCOAP_SERVER_TRANSPORT=ws python3 -m unittest tests.test_server.TestServerWS --failfast`.

aaugustin:
Interesting.

If you're talking of an individual connection, that is, a `WebSocketServerProtocol` instance, I would suspect the fact that connections have a logger which have a reference to the connection:

https://github.com/aaugustin/websockets/blob/10.0/src/websockets/legacy/protocol.py#L221

I'd be open to tweaking this with a `weakref` to facilitate garbage collection.

But I think you're talking of  the `.close()` and `.wait_closed()` methods of the `WebSocketServer` object. That doesn't ring a bell.

I'm not seeing any objects provided by websockets in the reference cycles visible in the CI logs. Do some of the classes involved inherit classes provided by websockets?

chrysn:
It's indeed something around the logging -- neutering the self.logger (by brutally assigning `self.logger = logging`) made my tests pass. Still looking for the precise cause for fix that's done with precision and not a butcher's knife.

Thanks for the pointer, I'll have a look whether a weakref or maybe something else around logging fixes things. (I've spent more time than I'd like to admit finding that lazy logs cause more trouble than they're worth, maybe it's something like that).

chrysn:
[...]

PR #1750, opened 2026-08-20
Break reference cycles keeping closed connections alive

### LINKED ISSUE #1749: Reference cycles keep closed connections alive until a full GC pass
Closed connections (asyncio, sync, and trio implementations alike) are never
freed by reference counting: two reference cycles keep every connection alive
until the cyclic garbage collector runs. On servers with many connections and
high connect/disconnect churn, this causes very large memory growth, because
CPython 3.13+ collects the oldest generation incrementally and lags far behind
the churn on large heaps.

Observed in production (websockets 15.0.1, CPython 3.14.4, ~450 concurrent
connections with heavy churn): a gc census on the live process showed 13,589
`ServerConnection` and 3,335 `SSLProtocol` instances alive with only 453
connections actually open. A forced `gc.collect()` freed ~1.75 GB — and took
16 seconds, a stop-the-world pause long enough to break the process's other
websocket connections via ping timeout.

Both cycles affect every connection — graceful close and abrupt abort alike.
Verified on 15.0.1 and reproduced on current main.

### Cycle 1: LoggerAdapter (all implementations)

`Connection.__init__` injects the connection into the protocol's logger:

```python
self.protocol.logger = logging.LoggerAdapter(
    self.protocol.logger,
    {"websocket": self},
)
```

connection → protocol → logger (LoggerAdapter) → `extra` dict → connection.
Nothing ever clears it. Disabling logging as documented (NullHandler /
propagate=False / setLevel) does not help — the cycle is created at
construction regardless of logging activity. The same pattern exists in
`asyncio/connection.py`, `sync/connection.py`, `trio/connection.py` (and
`legacy/protocol.py`, deprecated).

This was suspected once before in #1059, where @aaugustin noted: "I'd be open
to tweaking this with a `weakref` to facilitate garbage collection."

### Cycle 2: cancelled keepalive task (asyncio implementation)

`Connection.connection_lost()` cancels `self.keepalive_task` but keeps the
attribute. A cancelled `asyncio.Task` retains its `CancelledError`, whose
`__traceback__` holds the `Connection.keepalive()` coroutine frame, whose
locals include `self`:

connection → task → CancelledError → traceback → frame → connection.

### Reproduction

Prints `alive: True` on current main; only a `gc.collect()` frees the
connection:

```python
import asyncio, gc, weakref
from websockets.asyncio.client import connect
from websockets.asyncio.server import serve

async def handler(ws):
    try:
        async for _ in ws:
            pass
    except Exception:
        pass

async def main():
    gc.disable()
    gc.collect()
    server = await serve(handler, 'localhost', 8765)
    ws = await connect('ws://localhost:8765')
    await asyncio.sleep(0.2)
    ref = weakref.ref(next(iter(server.connections)))
    ws.transport.abort()          # also reproduces with: await ws.close()
    await asyncio.sleep(0.5)
    print('alive without gc:', ref() is not None)   # True == bug
    server.close()
    await server.wait_closed()

asyncio.run(main())
```

Break-set experiment on the dead connection (gc still disabled): clearing only
the logger cycle (`conn.protocol.logger.extra.clear()`) leaves the connection
alive; clearing the logger cycle *and* dropping the task
(`conn.keepalive_task = None`) makes the weakref die immediately. Both cycles
must be broken.

### Related

`recv_exc` stores arbitrary exceptions whose tracebacks can also pin frames
referencing the connection (e.g. exceptions raised inside `send_context()`).
This wasn't required to fix the common paths above, but may be worth
considering separately (`exc.__traceback__ = None` would keep the cause chain
intact for `close_exc.__cause__` while dropping the frames).

peteralm80:
Fixes #1749.

Closed connections could only be freed by the cyclic garbage collector, never
by reference counting, because of two reference cycles (details, production
numbers, and reproduction in the issue):

1. The `LoggerAdapter` that injects the `websocket` attribute into log records
   held a strong reference to the connection (all implementations).
2. In the asyncio implementation, the cancelled keepalive task retained a
   `CancelledError` whose traceback references the `keepalive()` frame and
   thus the connection.

Changes:

* New `ConnectionLoggerAdapter` in `websockets/utils.py` holds a weak
  reference to the connection and injects it into log records at logging time
  (the `weakref` approach suggested in #1059). Behavior is unchanged for any
  record emitted while the connection object is alive — including "connection
  handler failed" in the server, since the handler still references the
  connection at that point — because each record receives its own strong
  reference via `extra`. The documented adapter-wrapping pattern
  (`kwargs["extra"]["websocket"]` with `except KeyError`) keeps working.
  Used by the asyncio, sync, and trio implementations. The deprecated legacy
  implementation is left untouched.

* `Connection.connection_lost()` (asyncio) dereferences `keepalive_task` after
  cancelling it, so the task, its `CancelledError`, and the pinned frame die
  by refcount once the event loop finishes the cancellation.

Tests:

* Regression tests in `tests/asyncio/test_connection.py` (graceful close and
  abort) and `tests/sync/test_connection.py` assert with `gc.disable()` in
  effect that a `weakref` to a closed connection dies — i.e. the connection is
  freed by reference counting alone. They fail without the fix and are skipped
  on non-CPython implementations.
* Unit tests for `ConnectionLoggerAdapter` in `tests/test_utils.py`.
* Two existing keepalive tests that read `keepalive_task` after close now
  capture the task before closing.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/docs/project/changelog.rst
+++ b/docs/project/changelog.rst
@@ -32,6 +32,13 @@ notice.
 
 *In development*
 
+Bug fixes
+.........
+
+* Fixed reference cycles that delayed garbage collection of closed connections
+  until a pass of the cyclic garbage collector. Closed connections are now
+  freed immediately by reference counting.
+
 .. _17.0.1:
 
 17.0.1
--- a/src/websockets/asyncio/connection.py
+++ b/src/websockets/asyncio/connection.py
@@ -1019,6 +1019,13 @@ class Connection(asyncio.Protocol):
 
         self.set_recv_exc(exc)
 
+        # Clear the frames of recv_exc's traceback. Else, when recv_exc was
+        # raised in a method of this class e.g. data_received(), its traceback
+        # would keep the connection in a reference cycle. Frames finished
+        # running, so they may be cleared; formatting isn't affected.
+        if self.recv_exc is not None:
+            traceback.clear_frames(self.recv_exc.__traceback__)
+
         # Abort recv() and pending pings with a ConnectionClosed exception.
         self.recv_messages.close()
         self.terminate_pending_pings()
--- a/src/websockets/asyncio/connection.py
+++ b/src/websockets/asyncio/connection.py
@@ -3,7 +3,6 @@ from __future__ import annotations
 import asyncio
 import collections
 import contextlib
-import logging
 import random
 import struct
 import traceback
@@ -22,6 +21,7 @@ from ..frames import DATA_OPCODES, PONG, CloseCode, Frame
 from ..http11 import Request, Response
 from ..protocol import CLOSED, OPEN, Event, Protocol, State
 from ..typing import BytesLike, Data, DataLike, LoggerLike, Subprotocol
+from ..utils import ConnectionLoggerAdapter
 from .messages import Assembler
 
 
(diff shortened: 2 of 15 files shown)


gm122921980-create (approved):

gm122921980-create (approved):

gm122921980-create (approved):

gm122921980-create (approved):

gm122921980-create (approved):

aaugustin:
Superseded by #1752.

PR #1751, opened 2026-08-23
Adjust SSL certificate-verification error strings for AWS-LC compat

stewartsmith:
AWS-LC ssl.SSLCertVerificationError message differs from OpenSSL:
- `OpenSSL`:  "certificate verify failed: self signed certificate"
- `AWS-LC`:   "[SSL: CERTIFICATE_VERIFY_FAILED] CERTIFICATE_VERIFY_FAILED: self signed certificate (_ssl.c:1082)"

By adjusting the text being looked for to be the common substrings, the tests work when Python is built with the AWS-LC cryptographic library instead of OpenSSL.

### DIFF
--- a/tests/asyncio/test_client.py
+++ b/tests/asyncio/test_client.py
@@ -554,7 +554,7 @@ class SecureClientTests(unittest.IsolatedAsyncioTestCase):
                 async with connect(get_uri(server)):
                     self.fail("did not raise")
             self.assertIn(
-                "certificate verify failed: self signed certificate",
+                "self signed certificate",
                 str(raised.exception).replace("-", " "),
             )
 
@@ -568,7 +568,7 @@ class SecureClientTests(unittest.IsolatedAsyncioTestCase):
                 ):
                     self.fail("did not raise")
             self.assertIn(
-                "certificate verify failed: Hostname mismatch",
+                "Hostname mismatch",
                 str(raised.exception),
             )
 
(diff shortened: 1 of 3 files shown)


aaugustin:
Hello, I'm OK with the concept.

The same change must be applied in tests/trio/test_client.py, then I can merge the PR.

----

I assume that you are packaging websockets for AWS Linux and that you're attempting to run the tests as part of your packaging toolchain. I understand why you want to do this, but be aware that you aren't running tests properly: if you were, you would have noticed that you needed to change tests for the trio implementation.

This raises [questions](https://websockets.readthedocs.io/en/stable/project/contributing.html#packaging) on whether it makes sense for you to run the tests at all.

Here are a few options to run a decent subset of the test suite, from shortest to most explicit:

```
pip install tox
tox -e pyXY
```

```
pip install -r tests/requirements.txt
make
make tests
```

```
python -m pip install -r tests/requirements.txt
python setup.py build_ext --inplace
PYTHONPATH=src python -m unittest
```

stewartsmith:
> Hello, I'm OK with the concept.
> 
> The same change must be applied in tests/trio/test_client.py, then I can merge the PR.
> 
> I assume that you are packaging websockets for AWS Linux and that you're attempting to run the tests as part of your packaging toolchain. I understand why you want to do this, but be aware that you aren't running tests properly: if you were, you would have noticed that you needed to change tests for the trio implementation.

Yes, it's for building python with AWS-LC on Amazon Linux, along with a pile of modules. The packaging for `python-websockets` is near identical to current Fedora, so that's interesting. Current Fedora is also on version 15.0.1, which seems to be the key, as it seems like the `trio` code wasn't added until after then, and probably why all the tests passed there.

Now there's also the question of why tests passed for me... and I'll 100% blame the LLM answering the question of "how do I run tests for this" given my general ignorance of the Python packaging world.

Anyway, thanks for the pointers and for merging the PR!

aaugustin:
I'll look into committing LLM-oriented documentation that explains (among other things) how to run tests 👍 

Currently I have it only locally because the standards are not quite settled.

PR #1758, opened 2026-09-10
Fix non-ASCII headers in the legacy HTTP parser

thorvald-oai:
Preserve non-ASCII header values in the legacy HTTP parser by avoiding a second validation after surrogate decoding. Raw header validation remains in place.

Includes a regression test for a UTF-8 header value. This applies the modern parser's [existing fix](https://github.com/python-websockets/websockets/commit/399bb08ec9ffc170868cd3eec100628b60d16fae) to the legacy parser.


aaugustin:
I didn't change the behavior of the legacy parser on purpose because it's only used by people who don't maintain their software at this point and I didn't want to break it for them.

aaugustin:
Sorry I misunderstood what you were proposing. I thought you were proposing to change the encoding, while you're actually only avoiding the double validation.

aaugustin:
For my understanding, what is it that led you to propose this change?

Just trying to maintain some understanding of how websockets is used and what problems occur :-)

thorvald-oai:
Uvicorn still uses the legacy implementation in websockets when configured with ws="websockets", and a few of our internal services still use that configuration.

aaugustin:
Now I understand that it's actually a regression in 7cdb8c6, which landed in version 16.1, and causes the legacy implementation to reject non-ASCII headers that were previously accepted.

PR #1760, opened 2026-09-12
Document graceful server shutdown

### LINKED ISSUE #1756: Consider adding close_connections argument to serve
The `close` / `shutdown` / `aclose` methods of servers (in asyncio / threading / trio implementations) take a `close_connections` argument to control whether the server should close connections proactively (the default behavior) or wait for clients to disconnect by themselves (which could be arbitrarily long).

When using serve as a context manager, there is no way to control this behavior. An obvious option would consist in adding a `close_connections` argument to `serve`, then proxy it to `close` / `shutdown` / `aclose` in `__aexit__` / `__exit__` / `__aexit__`.

If we do this, for completeness, we should also add `close_code` and `close_reason`. This means adding three arguments to `serve`. I'm not convinced that it's a good trade-off.

An alternative would consist in documenting the recommended pattern for controlling this behavior. Currently, the trio implementation shows an example of using `aclose` which can easily be extended with some parameter. The asyncio and threading implementations don't show examples of using `close` and `shutdown`.

GruffElixir:
Closes #1756.

The FAQ now documents how to stop both asyncio and threading servers while keeping existing connections open. The threading example uses `Server.shutdown(close_connections=False)` and waits for the server thread to finish.

Validation: `sphinx-build -W -b dummy . _build/dummy` from `docs/`.


aaugustin:
I believe this is already documented in the API docs for the threading implementation.

I don't really want to start replicating every answer in the FAQ for all implementations.

Issue #1761, opened 2026-09-13
Proxy credentials in the proxy URL are not percent-decoded


oliver-zehentleitner-aigent:
### Description

`parse_proxy()` takes `username` and `password` from `urllib.parse.urlparse(...)` as they are, without percent-decoding them. A password that contains a reserved character has to be percent-encoded to be representable in the URL at all (`socks5://alice:p%40ss@host:1080` for the password `p@ss`), but websockets then sends the literal string `p%40ss` to the proxy: `connect_socks_proxy()` hands `proxy.username` / `proxy.password` to python-socks unchanged, and `connect_http_proxy()` base64-encodes them unchanged for `Proxy-Authorization`.

python-socks itself (`python_socks._helpers.parse_proxy_url`) applies `unquote()` to both, so the same URL works with python-socks directly and with other clients that build on it, but not through websockets' `proxy=` parameter. RFC 3986 section 3.2.1 defines userinfo as percent-encoded.

### Reproduction

```python
>>> from websockets.proxy import parse_proxy
>>> p = parse_proxy("socks5://alice:p%40ss@127.0.0.1:1080")
>>> p.password
'p%40ss'
```

Against a SOCKS5 proxy with user/password auth the connection is rejected (`ProxyError: failed to connect to SOCKS proxy`), against an HTTP CONNECT proxy with basic auth it is answered with 407.

### Environment

websockets 16.0, Python 3.13.5, python-socks 3.1.1, Linux x86_64.

### Expected

`parse_proxy()` percent-decodes `username` and `password` (e.g. `urllib.parse.unquote`), like `python-socks` and `requests` do for proxy URLs, so that credentials with `@`, `:`, `/` or `%` can be used.

aaugustin:
The same problem may well exist in `parse_uri`, for WebSocket URIs or IRIs, in addition to proxy URIs.

In order to be fully consistent, we must confirm which encoding is expected:

- in IRIs
- in URIs
- in HTTP requests
- when using a SOCKS proxy
- (anywhere else I might be missing)
- internally within websockets — this should "unicode, no encoding"

and convert appropriately at boundaries.

websockets has heuristics to guess if it's getting and URI or an IRI, and the change suggested here may interfere with those heuristics.

aaugustin:
Boundary expectations

- IRI userinfo: raw Unicode credentials remain Unicode after parsing.
- URI userinfo: percent escapes are decoded during parsing; %40 becomes @.
[...]

PR #1762, opened 2026-09-14
Stop parsing completed HTTP proxy responses.

Sakshamm-Goyal:
# Stop parsing completed HTTP proxy responses

When an HTTP proxy rejects CONNECT with 407 and closes the connection, `connect()` raises `InvalidProxyStatus` as expected, but the asyncio callback also raises `AttributeError` by resuming the completed response parser. EOF followed by connection loss can also feed EOF twice.

Treat the completed response future as terminal and share EOF handling. Seven tests cover success, rejection, malformed responses, cancellation, incomplete responses and trailing data. A local authenticated CONNECT tunnel reproduces the callback error before the fix and verifies echo/rejection without callback errors afterward.

Python 3.13 tests, full and per-module 100% coverage, Ruff and strict mypy pass. An initial synchronous reconnect timing failure passed on rerun with the repository CI timeout factor. Other Python/platform combinations weren't run locally.

Prepared and validated with an AI coding assistant. This is separate from the credential-decoding behavior in #1761.


aaugustin:
What problem is this actually causing in real life?

aaugustin:
> the asyncio callback also raises `AttributeError`

What is "the asyncio callback" in this context?

Sakshamm-Goyal:
The concrete effect I reproduced is an extra, uncaught event-loop exception during cleanup after a proxy rejects CONNECT with 407 (for example, incorrect proxy credentials). `connect()` still raises the correct `InvalidProxyStatus`; this fix doesn't make the rejected connection succeed. I don't have a production incident or evidence of data loss or a hang to claim.

By “the asyncio callback” I meant `_SelectorSocketTransport._call_connection_lost()`, which calls `HTTPProxyConnection.connection_lost()`. The traceback is:

```text
_SelectorSocketTransport._call_connection_lost()
  -> HTTPProxyConnection.connection_lost()
  -> HTTPProxyConnection.run_parser()
  -> response.status_code
AttributeError: 'NoneType' object has no attribute 'status_code'
```

The CONNECT response parser already completed when it reported 407. Cleanup resumes that exhausted generator, whose next `StopIteration` has no response value.

I re-ran the loopback integration on Python 3.13.11: a real local CONNECT proxy, a WebSocket echo server, and the public `connect(..., proxy=...)` API. Correct credentials echo successfully; incorrect credentials raise `InvalidProxyStatus` on both versions. With the pre-fix client at `acff86b`, the loop's exception handler additionally receives the `AttributeError` above. With this PR's head, it doesn't. So the demonstrated user-facing problem is a spurious asyncio error log / exception-handler invocation on an otherwise correctly handled proxy rejection, not a failed error return from `connect()`.

PR #1769, opened 2026-09-19
Preserve percent-encoded paths and queries in IRIs

jakezwang:
Preserve existing percent escapes in paths and queries when converting IRIs to URIs, as specified by [RFC 3987 §3.1](https://www.rfc-editor.org/rfc/rfc3987#section-3.1).

For example, `ws://example.com/a%20b?q=é` currently requests `/a%2520b?q=%C3%A9` instead of `/a%20b?q=%C3%A9`. Adding `%` to the safe set prevents this second encoding while retaining UTF-8 conversion of Unicode characters.

Adds regressions for Unicode in the hostname, path and query with existing escapes, plus a changelog entry.

Validation: full suite passed on Python 3.11 and 3.14 (2,585 tests each); URI/client tests passed on 3.12, 3.13 and 3.15. Ruff, strict mypy, Sphinx, package build and Twine checks passed. URI coverage is 100%; overall coverage is 99% with the same seven unexecuted legacy lines on pristine main.


aaugustin:
Hello, could you clarify how this PR accounts for [my spec](https://github.com/python-websockets/websockets/issues/1761#issuecomment-5655072453) of what needs to be done to solve the issue.

It seems that you've only solved a small subset. If everything else is actually OK, can you provide the evidence?

It seems that you've just thrown an AI at the problem without understanding it, not even noticing that someone else had made the same mistake before you. This is a lack of respect.

Please just don't throw an AI at my comment; I have an AI, and I'd rather run it myself than through you.

